PR #2, opened 2023-06-22
Don't panic when cargo contract is not installed.

JoshOrndorff:


pmikolajczyk41:
There's a clippy screaming, so I checked out to #4. Thanks for contribution!

PR #3, opened 2023-06-22
Fix `cargo contract build` command

JoshOrndorff:
I couldn't make the `build` command work. I even installed `cargo-contract` with `cargo +nightly install cargo-contract`.

image

To solve this error, I just had to change the command slightly. Please confirm whether this is the correct fix.


pmikolajczyk41:
right, at some point I had problems with `Command` and I fixed it with the previous solution. But your code works for me as well, so I guess it's more general. Moved to #4

PR #15, opened 2023-06-29
Use `Session` in CLI

pmikolajczyk41:
1. `Session` has now two APIs: for chaining (`with_*` and `*_and`) and singular actions
2. removed duplicated code from `drink-cli/executor` (core code was duplicating `Session`)
3. some mambo-jumbo for sharing `ContractMessageTranscoder` (it is not `Clone` nor `Copy`...)

### DIFF
--- a/drink/src/session.rs
+++ b/drink/src/session.rs
@@ -4,7 +4,9 @@ use frame_support::{dispatch::DispatchError, sp_runtime::AccountId32, weights::W
 use pallet_contracts_primitives::{ContractExecResult, ContractInstantiateResult};
 use thiserror::Error;
 
-use crate::{contract_api::ContractApi, Sandbox, DEFAULT_ACTOR, DEFAULT_GAS_LIMIT};
+use crate::{
+    chain_api::ChainApi, contract_api::ContractApi, Sandbox, DEFAULT_ACTOR, DEFAULT_GAS_LIMIT,
+};
 
 #[derive(Error, Debug)]
 pub enum SessionError {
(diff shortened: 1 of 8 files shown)


drink/src/session.rs
```diff
@@ -93,33 +124,58 @@ impl Session {
...
+    pub fn call_and(
+        mut self,
+        address: Option<AccountId32>,
+        message: &str,
+        args: &[String],
+    ) -> Result<Self, SessionError> {
+        self.call(address, message, args).map(|_| self)
+    }
+
+    pub fn call(
```
piotrMocz: This is not very important but actually for a public API I'd prefer two methods:
```rust
pub fn call(&mut self, message: &str, args: &[String])
```
and:
```rust
pub fn call_with_address(&mut self, address: AccountId32, message: &str, agrs: &[String])
```
And have them both call a non-public `call_internal` method with a signature just like `call` now.
Just thinking out loud :)

examples/flipper/lib.rs
```diff
@@ -67,12 +67,12 @@ mod tests {
 
     #[test]
     fn flipping() -> Result<(), Box<dyn Error>> {
-        let init_value = Session::new(Some(transcoder()))?
-            .deploy(bytes(), "new", &["true"], vec![])?
-            .call("flip", &[])?
-            .call("flip", &[])?
-            .call("flip", &[])?
-            .call("get", &[])?
+        let init_value = Session::new(Some(Rc::new(transcoder())))?
```
piotrMocz: This probably deserves its own wrapper to hide the gory details :)

piotrMocz (approved):
To the extent that I understand the code, the changes seem great.
I left some nitpicky comments but please feel free to ignore them :)

drink/src/session.rs
pmikolajczyk41: done :ok_hand:

examples/flipper/lib.rs
pmikolajczyk41: I moved them to the `transcoder` function - I hope that it will be soon hidden by the ink e2e framework

PR #16, opened 2023-06-29
Add makefile

pmikolajczyk41:


Makefile
```diff
@@ -0,0 +1,13 @@
+.PHONY: run build clean help
+
+run: ## Run the project
+	cargo run --release
```
deuszx, line 4: How does cargo know which binary to run here? Does it detect there's a singly `fn main` in the whole workspace?

pmikolajczyk41: in the workspace there's only one library and one binary

Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+build: ## Build the project
+	cargo build --release
+
+clean: ## Clean all the workspace build files
+	cargo clean
+
+help: ## Displays this help
+	@awk 'BEGIN {FS = ":.*##"; printf "Usage:\n  make \033[1;36m<target>\033[0m\n\nTargets:\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  \033[1;36m%-25s\033[0m %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
```
deuszx, line 17: You could add `cargo fmt` and `cargo clippy` but it's up to you.

deuszx (approved):

Makefile
pmikolajczyk41: added

PR #20, opened 2023-06-30
Add documentation for `drink`

pmikolajczyk41:


piotrMocz (approved):
Yay! In the future, we could add parameter descriptions for functions like `call` but all in all this is amazing!

PR #21, opened 2023-07-04
Use the contract-build crate instead of calling cargo contract

### LINKED ISSUE #18: Use `contract-build` crate instead of calling `cargo-contract` command
This would lift the requirement of pre-installed cc tool.

piotrMocz:
This code is written in an almost functional style which may not be to everyone's taste. I'm happy to either rewrite `deploy` and `call` to this style, or convert this one to the imperative style :)

closes: #18

### DIFF
--- a/drink-cli/src/executor/contract.rs
+++ b/drink-cli/src/executor/contract.rs
@@ -1,27 +1,54 @@
 use std::{env, fs, path::PathBuf, rc::Rc};
+use std::path::Path;
 
 use contract_transcode::ContractMessageTranscoder;
+use contract_build::{BuildMode, ExecuteArgs, ManifestPath, OptimizationPasses, Verbosity};
 
 use crate::app_state::{print::format_contract_action, AppState, Contract};
 
 pub fn build(app_state: &mut AppState) {
-    let Ok(output) = std::process::Command::new("cargo")
-        .arg("contract")
-        .arg("build")
-        .arg("--release")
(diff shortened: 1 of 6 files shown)


drink-cli/src/executor/contract.rs
```diff
@@ -1,26 +1,61 @@
...
+                write!(f, "Invalid manifest path {}: {}", manifest_path.display(), err),
+            BuildError::BuildFailed { err } =>
+                write!(f, "Contract build failed: {}", err),
+            BuildError::WasmNotGenerated =>
+                write!(f, "Wasm code artifact not generated"),
+            BuildError::InvalidDestPath { err } =>
+                write!(f, "Invalid destination bundle path: {}", err),
+        }
+    }
+}
```
pmikolajczyk41: I suggest using `thiserror` (you can see its usage in the `drink` crate)

pmikolajczyk41 (approved):
clean code finally here! I'm happy that I was right to torment you to join here :stuck_out_tongue: 

there's one very strange thing, i.e. when the contract has to be built, my UI shows:
image

i.e. it puts last line of stderr from contract-build in the user input... actually it's not tragic, so I wouldn't treat it as a blocker; if you don't find an easy way to get rid of it, then please merge and don't bother with it now

regarding making other code pieces more functional: you are more than welcome!

pmikolajczyk41:
closes #21

Issue #8, opened 2023-06-23
Tab completing for the `cd` command


pmikolajczyk41:
> or if this is too hard, as a workaround, allow passing a starting directory like ./target/release/drink-cli exmaples/flipper

cc: @JoshOrndorff

pmikolajczyk41:
I've spent some time trying to get this working, but actually, I don't see any easy, good working solution. `clap_complete` is able to generate a short shell script, but it's still pretty troublesome to run it (it's not a single command). Also, displaying available completions is a question mark :|

While I agree that it would be a nice feature, for now we will have to park it.

piotrMocz:
@JoshOrndorff we merged the change allowing to pass a starting directory, you can give that a spin! ;)

PR #23, opened 2023-07-05
Allow starting the CLI with an initial pwd set

piotrMocz:
This was mentioned as a half-measure for #8 and it seems to cover a lot of the use cases (assuming users start `drink-cli` to mostly work on one contract.

Usage:
```
$ drink-cli --help
Usage: drink-cli [OPTIONS]

Options:
  -p, --path <DIRECTORY>  Starts the CLI in the provided directory
  -h, --help              Print help
  -V, --version           Print version
```

So you're able to start it like:
```
$ drink-cli examples/flipper
```
And immediately call `build`.

### DIFF
--- a/drink-cli/src/app_state/mod.rs
+++ b/drink-cli/src/app_state/mod.rs
@@ -47,6 +47,17 @@ pub struct UiState {
     pub show_help: bool,
 }
 
+impl UiState {
+    pub fn new(pwd_override: Option<PathBuf>) -> Self {
+        let pwd = pwd_override.unwrap_or(
+            env::current_dir().expect("Failed to get current directory"));
+        UiState {
+            pwd,
+            ..Default::default()
+        }
+    }
+}
+
 impl Default for UiState {
     fn default() -> Self {
         UiState {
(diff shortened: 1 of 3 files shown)


drink-cli/src/app_state/mod.rs
```diff
@@ -47,6 +47,17 @@ pub struct UiState {
     pub show_help: bool,
 }
 
+impl UiState {
+    pub fn new(pwd_override: Option<PathBuf>) -> Self {
+        let pwd = pwd_override.unwrap_or(
+            env::current_dir().expect("Failed to get current directory"));
+        UiState {
+            pwd,
+            ..Default::default()
```
pmikolajczyk41: in order to deduplicate a bit, I would specify here the rest of the arguments and in the `default` implementation I would put `Self::new(None)`

pmikolajczyk41 (changes requested):
when I run: `cargo run --release -- -p /home/<my local path>/drink/examples/flipper`, optionally build the contract and try to deploy it, I get error about finding contract file

this helps:
```rust
fn find_wasm_blob(pwd: &PathBuf) -> Option<(String, PathBuf)> {
    let Ok(entries) = fs::read_dir(pwd.join("target/ink")) else {
```

drink-cli/src/app_state/mod.rs
piotrMocz: Yeah that's actually a way more natural order of dependencies (i.e. the constructor doesn't assume the existence of the `Default` impl, which kind of seemed like an anti-pattern anyway... Thanks!

pmikolajczyk41 (approved):

Issue #22, opened 2023-07-04
Make `drink` generic over runtime


pmikolajczyk41:
`Sandbox` should be generic over runtime. We should also provide a few example instantiations. E.g.:
 - `minimal` - the current one
 - `substrate-contracts-node` - taken from `substrate-contracts-node`
 - `Aleph` - taken from `aleph-node`

pmikolajczyk41:
on it

pmikolajczyk41:
for now we can't do more than #24  - both aleph and substrate-contracts-node runtimes are not publishable and thus we can't use them here (since git-dependencies would prevent us from publishing drink)

PR #24, opened 2023-07-06
Make `drink` generic over runtime

pmikolajczyk41:
1. We abstract `Sandbox` and `Session` to be generic over underlying runtime.
2. We provide the `Runtime` trait that captures minimal requirements (three necessary pallets, few basic types). Some of them might be relaxed further, but it's not necessary now and would introduce denser type-soup.
3. We provide implementation for `MinimalRuntime` and instantiate `drink-cli` with it

### DIFF
--- a/drink-cli/src/executor/mod.rs
+++ b/drink-cli/src/executor/mod.rs
@@ -68,10 +68,12 @@ pub fn execute(app_state: &mut AppState) -> Result<()> {
 
 fn build_blocks(app_state: &mut AppState, count: u64) {
     for _ in 0..count {
-        app_state.session.chain_api().build_block();
+        app_state.chain_info.block_height = app_state
+            .session
+            .chain_api()
+            .build_block()
+            .expect("Failed to build block - chain is broken");
     }
-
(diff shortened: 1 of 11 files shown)


pmikolajczyk41:
mostly solves #22

drink-cli/src/executor/mod.rs
```diff
@@ -68,11 +68,13 @@ pub fn execute(app_state: &mut AppState) -> Result<()> {
 
 fn build_blocks(app_state: &mut AppState, count: u64) {
     for _ in 0..count {
-        app_state.session.chain_api().build_block();
+        app_state.chain_info.block_height = app_state
```
piotrMocz: Any reason why we're not using the `ChainApi::build_block` method?

piotrMocz (approved):
LGTM :+1:

drink-cli/src/executor/mod.rs
pmikolajczyk41: nice spot, changed to `build_blocks`

PR #38, opened 2023-08-18
Minor changes required for ink! integration

pmikolajczyk41:


drink/src/contract_api.rs
```diff
@@ -2,47 +2,64 @@
...
 use crate::{runtime::Runtime, EventRecordOf, Sandbox};
 
 /// Interface for contract-related operations.
 pub trait ContractApi<R: Runtime> {
     /// Interface for `bare_instantiate` contract call.
+    #[allow(clippy::too_many_arguments)]
     fn deploy_contract(
         &mut self,
         contract_bytes: Vec<u8>,
+        value: u128,
```
piotrMocz, line 28: Maybe a word of explanation of what the `value` here actually is? A tiny docstring ;)
Or you could add it to the top-level docstring of the module, saying sth like "the `value` argument of calls in this module is the amount of the native currency transferred to the contract alongside the call".
Makes sense?

piotrMocz (approved):
LGTM. I'm a sucker for docs so just one tiny comment ;)

drink/src/contract_api.rs
pmikolajczyk41: added docs

PR #39, opened 2023-08-24
Dry running

### LINKED ISSUE #37: Allow dry-running
Enrich `ContractsApi` with a possibility to dry-run instantiation or calling.

# Possible approach
A possible way of approaching it is to copy storage before, running normal procedure within externalities and then restoring old storage.

pmikolajczyk41:
Closes #37 

Relevant problems: https://github.com/paritytech/polkadot-sdk/issues/123

### DIFF
--- a/drink/src/chain_api.rs
+++ b/drink/src/chain_api.rs
@@ -39,6 +39,13 @@ pub trait ChainApi {
     ///
     /// * `address` - The address of the account to query.
     fn balance(&mut self, address: &AccountId32) -> u128;
+
+    /// Run an action without modifying the storage.
+    ///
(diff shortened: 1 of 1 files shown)


piotrMocz (approved):
Code looks very good, although technically this is not a dry run, I think — e.g. if the `commit_all` call fails, it will leave the state changed, won't it?

pmikolajczyk41:
it will panic, yes, but it is the best we can do now (at least this is my conclusion after 2h digging in the Substrate code)... I requested Parity to provide a true dry-run mechanism, and once it is there, I will replace it here

PR #40, opened 2023-08-25
Call runtime

### LINKED ISSUE #36: Allow arbitrary runtime call
Enrich `ChainApi` with a similar method to https://github.com/paritytech/ink/blob/cfada8405c212745e183205f2aa73c12200d4c1a/crates/e2e/src/backend.rs#L74

It might require a macro that combines `pallet_` with the `pallet` string argument...

pmikolajczyk41:
closes #36 

While the current solution is quite clean, its usage might turn out to be onerous (the vanilla way is to have a target pallet as a dependency). However, one can always prepare the object call manually without pallet (e.g. with subxt), encode it and then decode it :shrug: 

This problem will rise in ink! integration, where we have to implement:
```rust 
fn runtime_call<'a>(
        &mut self,
        _origin: &Keypair,
        _pallet_name: &'a str,
        _call_name: &'a str,
        _call_data: Vec<Value>,
    )
```

where `Value` comes from `subxt`. For this however, I will try to change the API there a bit (it was designed for subxt only). Then, integration with this new code should be much easier.

### DIFF
--- a/drink/src/chain_api.rs
+++ b/drink/src/chain_api.rs
@@ -57,6 +57,7 @@ pub trait ChainApi<R: Runtime> {
     /// # Arguments
     ///
     /// * `call` - The runtime call to execute.
+    /// * `origin` - The origin of the call.
     fn runtime_call(
         &mut self,
         call: RuntimeCall<R>,
--- a/drink/src/chain_api.rs
+++ b/drink/src/chain_api.rs
@@ -1,11 +1,16 @@
 //! Basic chain API.
 
+use frame_support::dispatch::Dispatchable;
+use frame_support::sp_runtime::DispatchResultWithInfo;
 use frame_support::{sp_runtime::AccountId32, traits::tokens::currency::Currency};
 
 use crate::{DrinkResult, Error, Runtime, Sandbox};
 
+/// The runtime call type for a particular runtime.
+pub type RuntimeCall<R> = <R as frame_system::Config>::RuntimeCall;
+
 /// Interface for basic chain operations.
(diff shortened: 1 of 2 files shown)


drink/src/chain_api.rs
```diff
@@ -109,4 +133,25 @@ mod tests {
 
         assert_eq!(sandbox.balance(&DEFAULT_ACTOR), initial_balance);
     }
+
+    #[test]
+    fn runtime_call_works() {
```
deuszx, line 139: Can you also include a test where the call fails due to a non-existing pallet?

drink/src/chain_api.rs
```diff
@@ -46,9 +51,20 @@ pub trait ChainApi {
     ///
     /// * `action` - The action to run.
     fn dry_run<T>(&mut self, action: impl FnOnce(&mut Self) -> T) -> T;
+
+    /// Execute a runtime call (dispatchable).
+    ///
+    /// # Arguments
+    ///
+    /// * `call` - The runtime call to execute.
```
deuszx, line 59: and `origin` ;)

deuszx (approved):
no blockers

drink/src/chain_api.rs
pmikolajczyk41: I can't - the method expects a valid `RuntimeCall` object

drink/src/chain_api.rs
pmikolajczyk41: added :white_check_mark:

PR #42, opened 2023-08-28
Contract API tests

pmikolajczyk41:
Adding tests for ContractAPI. Also, bump version to 1.3 - it should be now ready to fill the gaps in ink!


drink/Cargo.toml
```diff
@@ -21,5 +21,8 @@ parity-scale-codec = { workspace = true }
 scale-info = { workspace = true }
 thiserror = { workspace = true }
 
+[dev-dependencies]
+wat = { workspace = true }
```
deuszx, line 25: Wat is das?

pmikolajczyk41: can turn wat to a wasm blob :man_shrugging:

deuszx: Why did you have to write a wat by hand?

pmikolajczyk41: 1. pallet-contracts' unit tests also use wat instead of full Rust contract compiled to wasm
2. no need for compiling contract to wasm / no need to keep obscure bytes
3. kinda readable
4. you don't have to encode any selectors or arguments: `data: vec![]` works well for calling the only method

drink/test-resources/transfer.wat
```diff
@@ -0,0 +1,44 @@
...
+
+	;; [40, 44) here we store the return code of the transfer
+
+	(func (export "deploy"))
+
+	(func (export "call")
+		(i32.store
+			(i32.const 40)
+			(call $seal_transfer
+				(i32.const 0) ;; ptr to destination address
```
deuszx: Using zero-address is incorrect as the private key for it is well-known ;P 

Doesn't matter in this context of course xD

pmikolajczyk41: copied from Substrate :P

drink/src/contract_api.rs
```diff
@@ -140,3 +140,121 @@ pub fn decode_debug_buffer(buffer: &[u8]) -> Vec<String> {
...
+        ]
+        .concat();
+        wat::parse_file(path).expect("Failed to parse wat file")
+    }
+
+    #[test]
+    fn can_upload_code() {
+        let mut sandbox = Sandbox::<MinimalRuntime>::new().unwrap();
+        let wasm_binary = compile_module("transfer");
+        let hash = <<MinimalRuntime as frame_system::Config>::Hashing>::hash(&wasm_binary);
```
deuszx, line 172: <3

drink/test-resources/transfer.wat
deuszx: good practices etc :P

pmikolajczyk41: removed transferring at all (it was silently failing due to insufficient contract's balance)

deuszx: and tests were passing? How's that po(a)ssible? ie how was the event emitted if the txn failed? Or at this level the event emition is not yet affecting what ends up in the final block?

deuszx (approved):

drink/test-resources/transfer.wat
pmikolajczyk41: okay, taking that back - I have no idea why it was passing

PR #45, opened 2023-09-04
Cross-contract call tracing

pmikolajczyk41:
We make use of the recent extension to pallet-contracts, which enables tracing cross-contract calls.

### DIFF
--- a/drink/src/lib.rs
+++ b/drink/src/lib.rs
@@ -17,6 +17,7 @@ pub use frame_support::{sp_runtime::AccountId32, weights::Weight};
 use frame_system::{EventRecord, GenesisConfig};
 
 use crate::pallet_contracts_debugging::DebugExt;
+use crate::runtime::pallet_contracts_debugging::NoopDebugExt;
 use crate::runtime::*;
 
 /// Main result type for the drink crate.
@@ -63,10 +64,12 @@ impl<R: Runtime> Sandbox<R> {
             .execute_with(|| R::initialize_block(1, Default::default()))
             .map_err(Error::BlockInitialize)?;
 
+        sandbox.override_debug_handle(DebugExt(Box::new(NoopDebugExt {})));
+
         Ok(sandbox)
     }
 
-    pub fn register_debug_handle(&mut self, d: DebugExt) {
+    pub fn override_debug_handle(&mut self, d: DebugExt) {
         self.externalities.register_extension(d);
     }
 }
--- a/drink/src/lib.rs
+++ b/drink/src/lib.rs
@@ -64,11 +64,16 @@ impl<R: Runtime> Sandbox<R> {
             .execute_with(|| R::initialize_block(1, Default::default()))
             .map_err(Error::BlockInitialize)?;
 
+        // We register a noop debug extension by default.
         sandbox.override_debug_handle(DebugExt(Box::new(NoopDebugExt {})));
 
         Ok(sandbox)
     }
 
+    /// Overrides the debug extension.
+    ///
+    /// By default, a new `Sandbox` instance is created with a noop debug extension. This method
+    /// allows to override it with a custom debug extension.
     pub fn override_debug_handle(&mut self, d: DebugExt) {
         self.externalities.register_extension(d);
     }
--- a/drink/src/lib.rs
+++ b/drink/src/lib.rs
@@ -16,6 +16,7 @@ use frame_support::{sp_io::TestExternalities, sp_runtime::BuildStorage};
 pub use frame_support::{sp_runtime::AccountId32, weights::Weight};
 use frame_system::{EventRecord, GenesisConfig};
 
+use crate::pallet_contracts_debugging::DebugExt;
 use crate::runtime::*;
 
 /// Main result type for the drink crate.
@@ -63,5 +64,9 @@ impl<R: Runtime> Sandbox<R> {
             .map_err(Error::BlockInitialize)?;
 
         Ok(sandbox)
+    }
+
+    pub fn register_debug_handle(&mut self, d: DebugExt) {
+        self.externalities.register_extension(d);
     }
 }
--- a/drink/src/runtime/minimal.rs
+++ b/drink/src/runtime/minimal.rs
@@ -1,22 +1,25 @@
 #![allow(missing_docs)] // `construct_macro` doesn't allow doc comments for the runtime type.
 
+use std::time::SystemTime;
+
 use frame_support::{
     parameter_types,
     sp_runtime::{
         testing::H256,
         traits::{BlakeTwo256, Convert, IdentityLookup},
-        AccountId32, BuildStorage,
+        AccountId32, BuildStorage, Storage,
     },
-    traits::{ConstBool, ConstU128, ConstU32, ConstU64, Currency, Randomness},
+    traits::{ConstBool, ConstU128, ConstU32, ConstU64, Currency, Hooks, Randomness},
     weights::Weight,
 };
-use pallet_contracts::{DefaultAddressGenerator, Frame, Schedule};
-
 // Re-export all pallets.
 pub use frame_system;
 pub use pallet_balances;
 pub use pallet_contracts;
+use pallet_contracts::{DefaultAddressGenerator, Frame, Schedule};
 pub use pallet_timestamp;
+
+use crate::{runtime::pallet_contracts_debugging::DrinkDebug, Runtime, DEFAULT_ACTOR};
 
 type Block = frame_system::mocking::MockBlock<MinimalRuntime>;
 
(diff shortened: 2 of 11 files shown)


drink/src/lib.rs
```diff
@@ -62,6 +64,17 @@ impl<R: Runtime> Sandbox<R> {
             .execute_with(|| R::initialize_block(1, Default::default()))
             .map_err(Error::BlockInitialize)?;
 
+        // We register a noop debug extension by default.
+        sandbox.override_debug_handle(DebugExt(Box::new(NoopDebugExt {})));
```
deuszx, line 68: What would happen if you didn't explicitly register a `noop` extension?

drink/src/runtime/minimal.rs
```diff
@@ -121,15 +124,9 @@ impl pallet_contracts::Config for MinimalRuntime {
     type MaxDebugBufferLen = ConstU32<{ 2 * 1024 * 1024 }>;
     type Migrations = ();
     type DefaultDepositLimit = DefaultDepositLimit;
-    type Debug = ();
+    type Debug = DrinkDebug;
```
deuszx, line 127: Should that become the default for `minimal` runtime? I'd expect that the default for it would be `()` (or `NoopDebugExt`).

drink/src/runtime/pallet_contracts_debugging.rs
```diff
@@ -0,0 +1,114 @@
...
+pub trait DebugExtT {
+    /// Called after a contract call is made.
+    fn after_call(
+        &self,
+        _contract_address: Vec<u8>,
+        _is_call: bool,
+        _input_data: Vec<u8>,
+        _result: Vec<u8>,
+    ) {
+    }
```
deuszx, line 44: No `before_call`?

examples/cross-contract-call-tracing/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    thread_local! {
+        static OUTER_ADDRESS:  RefCell<Option<AccountId32>> = RefCell::new(None);
+        static MIDDLE_ADDRESS:  RefCell<Option<AccountId32>> = RefCell::new(None);
+        static INNER_ADDRESS:  RefCell<Option<AccountId32>> = RefCell::new(None);
+    }
+
+    struct TestDebugger;
+    impl DebugExtT for TestDebugger {
+        fn after_call(
```
deuszx, line 102: I'd prefer to see a test where `DebugExtT::after_call` is used to extract some data from the `after_call` trace (and `before_call` as well) which is then used to make some assertions in the unit test itself. I think an example like that would be far more useful for future users of this tracing feature.

deuszx (commented):
Left a couple of comments. In general, code looks confusing (which in Substrate usually means _good_) but the example usage (test) is not very useful.

drink/src/runtime/pallet_contracts_debugging.rs
pmikolajczyk41: due to API change here https://github.com/paritytech/substrate/pull/14789 it was easier to skip `before_call` for now; I will add it as a follow up (to avoid increasing complexity of this PR)

drink/src/runtime/minimal.rs
pmikolajczyk41: since runtime cannot be generic or customizable, I had to push the 'defaultness' down to `NoopDebugExt`

in other words:
 - I want to have just one runtime type in drink (`MinimalRuntime`); not counting things like `AlephRuntime` or `SubstrateContractsNodeRuntime` which actaully are just reexports
 - I want to give users choice if they want noop or custom logic

thus I have to put here `DrinkDebug` and push potential customization to extensions that are accessed by `DrinkDebug`

drink/src/lib.rs
pmikolajczyk41: Then we would get `"Failed to find 'DebugExt' extension"` error with the first contract instantiation / call.

examples/cross-contract-call-tracing/lib.rs
pmikolajczyk41: this would require interpreting `Value`s from `input_data` :|

drink/src/lib.rs
deuszx: So registering extensions is in whose interest? Runtime provider?

pmikolajczyk41: When you create a `Sandbox` or `Session` there is already a noop extension registered. If you want your own, you just override it

DamianStraszak (approved):

examples/cross-contract-call-tracing/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn bytes() -> Vec<u8> {
+        let path = "target/ink/cross_contract_call_tracing.wasm";
+        fs::read(path).expect("Failed to find or read contract file")
+    }
+
+    fn ok(v: Value) -> Value {
+        Value::Tuple(Tuple::new(Some("Ok"), vec![v]))
+    }
+
+    thread_local! {
```
DamianStraszak, line 94: Can't we put these inside of `TestDebugger`?

examples/cross-contract-call-tracing/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+        static OUTER_ADDRESS:  RefCell<Option<AccountId32>> = RefCell::new(None);
+        static MIDDLE_ADDRESS:  RefCell<Option<AccountId32>> = RefCell::new(None);
+        static INNER_ADDRESS:  RefCell<Option<AccountId32>> = RefCell::new(None);
```
DamianStraszak, line 97: Can't we put these 3 inside of `TestDebugger`?

DamianStraszak: Oh, I see, we can't because at the moment of creating it, they are not yet deployed?

pmikolajczyk41: yes, exactly; but we could put `TestDebugger` inside the testcase, but this makes things very opaque

drink/src/runtime/pallet_contracts_debugging.rs
```diff
@@ -0,0 +1,114 @@
...
+//! The interaction between end-user and runtime is as follows:
+//! 1. At some points during execution, the pallet invokes some callback through its configuration
+//! parameter `Debug`.
+//! 2. In order to forward the callback outside the runtime, `Debug` will call a runtime interface,
+//! that will then forward the call further to the proper runtime extension.
+//! 3. The runtime extension can be fully controlled by the end-user. It just has to be registered
+//! in the runtime.
+//!
+//! So, in brief: pallet-contracts -> runtime interface -> runtime extension
+//!              |<-----------runtime side-------------->|<---user side--->|
```
deuszx, line 15: Is runtime extension really on the "user" side? I guess it depends who the user here is - I'd say it's _drink!_ and not drink's user? Or more generally - user is where the set up of the runtime happens. Here that's drink!'s `Sandbox` and in normal chain that would be a node.

Is the above correct?

drink/src/runtime/pallet_contracts_debugging.rs
```diff
@@ -0,0 +1,114 @@
...
+//!
+//! # Passing objects between runtime and runtime extension
+//!
+//! Unfortunately, runtime interface that lies between runtime and the end-user accepts only
+//! very simlpe argument types and those that implement some specific traits. This means that
```
deuszx:
```suggestion
//! very simple argument types and those that implement some specific traits. This means that
```

drink/src/runtime/pallet_contracts_debugging.rs
```diff
@@ -0,0 +1,114 @@
...
+//! usually, complext objects will be passed in their encoded form (`Vec<u8>` obtained with scale
```
deuszx:
```suggestion
//! usually, complex objects will be passed in their encoded form (`Vec<u8>` obtained with scale
```

drink/src/runtime/minimal.rs
deuszx: Why keep the `NoopDebugExt` around then?

drink/src/runtime/pallet_contracts_debugging.rs
pmikolajczyk41: not necessary - any drink! user can provide their own extension that would express their own intention (and possible could even capture some local environment)

deuszx (approved):

drink/src/runtime/minimal.rs
deuszx: Explained off-gh.

drink/src/runtime/pallet_contracts_debugging.rs
pmikolajczyk41: meh, my IDE stopped spellchecking for some reason :| fixed

drink/src/runtime/pallet_contracts_debugging.rs
pmikolajczyk41: fixed

PR #50, opened 2023-09-08
Stop returning `Value` from `Session`. Return good old scale encoded values

pmikolajczyk41:
`Value` is an awful thing to return. Hard to create, obscure, bleh. Go with encoded bytes.

### DIFF
--- a/drink/src/session.rs
+++ b/drink/src/session.rs
@@ -364,7 +364,7 @@ impl<R: Runtime> Session<R> {
         self.call_returns.last().cloned()
     }
 
-    /// Returns the last value (in the encoded form) returned from calling a contract.
+    /// Returns the last value (in the decoded form) returned from calling a contract.
     ///
     /// Returns `None` if there has been no call yet, or if decoding failed.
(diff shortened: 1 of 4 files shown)


drink/src/session.rs
```diff
@@ -336,11 +357,21 @@ impl<R: Runtime> Session<R> {
...
-    /// Returns the last value returned from calling a contract.
-    pub fn last_call_return(&self) -> Option<Value> {
+    /// Returns the last value (in the encoded form) returned from calling a contract.
+    ///
+    /// Returns `None` if there has been no call yet.
+    pub fn last_call_raw_return(&self) -> Option<Vec<u8>> {
         self.call_returns.last().cloned()
     }
 
+    /// Returns the last value (in the encoded form) returned from calling a contract.
```
piotrMocz: Shouldn’t the docstring say „decoded” here?

piotrMocz (approved):

drink/src/session.rs
pmikolajczyk41: fixed

PR #52, opened 2023-09-19
Allow endowing contract while deploying or calling

pmikolajczyk41:
Enrich session API with optional endowment argument. Additionally, we add to CI running checks on the recent example (for cross-contract call tracing).


piotrMocz (approved):
LGTM — the tests are failing because of the missing `endowment` argument to some of the `call` methods but that should be a trivial fix

PR #56, opened 2023-09-21
Automatic publishing

pmikolajczyk41:
We add CI workflow for automated publishing crates in the workspace. In PR, only dry run will be verified. On push to main, publish.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,26 @@
+---
+name: Publish to crates.io
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+concurrency:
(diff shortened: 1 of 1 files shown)


obrok (approved):
I'm OK with merging as-is if you're not interested in the argument about `Cargo.lock` that @deuszx made on slack. Did you add the CRATES_IO_TOKEN var to repo settings?

pmikolajczyk41:
regarding variable - yes, I did, thanks for checking

regarding Cargo.lock - I remember your troubles, but in drink case it's rather not a problem; actually, only maintained and used published thing from this repository is library (so no `cargo install` problems); `drink-cli` is not actively developed, so I wouldn't take care of it that edge cases now; also, you were using v1 of this GH action, there's a chance that things got better in v2

PR #58, opened 2023-09-27
Session: reorganize transcoder management

pmikolajczyk41:
Instead of keeping only a single 'active' transcoder, `Session` object now keeps a transcoder registry, i.e. a mapping from contract address to a corresponding transcoder. The new API:
 - the `Session` constructor takes no arguments
 - while deploying a contract you must pass a shared reference to a corresponding transcoder; in case of success, the new contract address will be internally connected with the transcoder
 - while calling a contract address, the corresponding transcoder will be looked up in the registry
 - for already deployed contracts, you can register their transcoders with `set_transcoder` or `with_transcoder` methods

Also, a minor refactor has been done: error and result types related to session have been moved to a submodule

### DIFF
--- a/drink/src/session.rs
+++ b/drink/src/session.rs
@@ -4,10 +4,9 @@ use std::{fmt::Debug, mem, rc::Rc};
 
 pub use contract_transcode;
 use contract_transcode::ContractMessageTranscoder;
-use frame_support::{sp_runtime::DispatchError, weights::Weight};
+use frame_support::{ weights::Weight};
 use pallet_contracts_primitives::{ContractExecResult, ContractInstantiateResult};
 use parity_scale_codec::Decode;
-use thiserror::Error;
 
 use crate::{
     chain_api::ChainApi,
@@ -16,6 +15,9 @@ use crate::{
     runtime::{AccountIdFor, HashFor, Runtime},
     EventRecordOf, Sandbox, DEFAULT_GAS_LIMIT,
 };
+
+pub mod errors;
+use errors::{SessionError, MessageResult};
 
 type Balance = u128;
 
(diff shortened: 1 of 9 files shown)


drink/src/session.rs
```diff
@@ -266,6 +223,9 @@ impl<R: Runtime> Session<R> {
             Ok(exec_result) => {
                 let address = exec_result.account_id.clone();
                 self.deploy_returns.push(address.clone());
+
+                self.transcoders.register(address.clone(), transcoder);
```
deuszx, line 227: nitpick: transcoders can be registered per code-hash. Not necessary to do that rn.

pmikolajczyk41: but when calling a contract, I don't know its code-hash right away, do I? it would require to query the storage, right?

deuszx (approved):

drink/src/session.rs
deuszx: I think when you deploy you could register the code hash.

deuszx: But this matters only for cases when:
1. You upload code.
2. Call `new` multiple times
3. All of the instances share the same transcoder.

pmikolajczyk41: I know, but the problem is when you call a contract and you have to get the transcoder for its code-hash knowing only its address

deuszx: It would be a problem if you didn't deploy/instantiate that contract yourself but I don't think that's the case with drink? All contracts/addresses one will be interacting with are going to be created via this _session_ object, right? So when calling `deploy` or `instantiate` we can record the address and associate it with the code hash. You will need a separate map `Mapping<AccountId, Contract>`.

pmikolajczyk41: but then, there's no need for code-hash lookup, since every contract must have been already deployed with transcoder

drink/src/session/transcoding.rs
```diff
@@ -0,0 +1,23 @@
...
+        Self {
+            transcoders: BTreeMap::new(),
+        }
+    }
+
+    pub fn register(&mut self, contract: Contract, transcoder: &Rc<ContractMessageTranscoder>) {
+        self.transcoders.insert(contract, Rc::clone(transcoder));
+    }
+
+    pub fn get(&self, contract: &Contract) -> Option<Rc<ContractMessageTranscoder>> {
```
piotrMocz, line 20: Would a `remove` also make sense?

examples/cross-contract-call-tracing/lib.rs
```diff
@@ -142,14 +142,17 @@ mod tests {
 
     #[test]
     fn test() -> Result<(), Box<dyn Error>> {
-        let mut session = Session::<MinimalRuntime>::new(Some(transcoder()))?;
+        let mut session = Session::<MinimalRuntime>::new()?;
         session.override_debug_handle(DebugExt(Box::new(TestDebugger {})));
 
-        let outer_address = session.deploy(bytes(), "new", NO_ARGS, vec![1], None)?;
+        let outer_address =
+            session.deploy(bytes(), "new", NO_ARGS, vec![1], None, &transcoder())?;
```
piotrMocz, line 149: I'm wondering, for purely aesthetic reasons, whether the optional argument should come last.

piotrMocz (approved):
LGTM, two questions but feel free to ignore as they are nitpicky

drink/src/session/transcoding.rs
pmikolajczyk41: actually I can't imagine a situation for that... even if you terminate a contract, there's no way you can deploy another contract with different metadata to the same address later

examples/cross-contract-call-tracing/lib.rs
pmikolajczyk41: actually in rust there are no optional arguments, just a name coincidence :P I was grouping these arguments by the context they are used; not sure if that's the most convenient order - I can make transcoder as the first argument, which will accidentally make storage deposit coming last :P

PR #61, opened 2023-10-05
Basic contract mocking

pmikolajczyk41:
Finally we introduce mocking feature. For now, user can deploy a mock contract with arbitrary number of mocked messages (defined with typed closures), like this:
```rust
let mocked_contract = ContractMock::new()
    .with_message([0,0,0,0], mock_message(|()| 41))
    .with_message([0,0,0,1], mock_message(|(account, value)| Ok(()));
let mock_address = session.mocking_api().deploy_mock(mocked_contract);
```

## Future upgrades
 - mock spies (registering calls for every mocked message)
 - defining custom fallback in case selector mismatch (panic, catch-all default, ignoring mock)
 - mocking existing contracts (mocking a single message of an already deployed contract)
 - ...

## Notes for reviewers
I made some minor refactors which unfortunately resulted in a bigger diff. The code regarding errors and tracing was just moved to another modules, so you can safely skip it.

The area of runtime interfaces and extensions is very hard, also for me, so if you have any troubles, please check the brief introduction in the docsting in `runtime/pallet_contracts_debugging.rs`.


piotrMocz (approved):
LGTM

.github/workflows/rust-checks.yml
```diff
@@ -59,6 +59,8 @@ jobs:
       - name: Run tests for examples
         shell: bash
         run: |
+          # todo: use loop xD
```
deuszx, line 62: You _could_ reimplement it in the `Makefile` and just call here. See [this](https://github.com/Cardinal-Cryptography/common-amm/blob/main/Makefile#L37-L67) for inspiration.

drink/src/lib.rs
```diff
@@ -77,7 +96,65 @@ impl<R: Runtime> Sandbox<R> {
...
+}
+
+impl<AccountId: Ord + Decode> InterceptingExtT for MockingExtension<AccountId> {
+    fn intercept_call(
+        &self,
+        contract_address: Vec<u8>,
+        _is_call: bool,
+        input_data: Vec<u8>,
+    ) -> Vec<u8> {
+        let contract_address = Decode::decode(&mut &contract_address[..])
```
deuszx, line 129: Would `contract_address.as_mut_slice()` work? https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_mut_slice

examples/mocking/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+version = "0.1.0"
+
+[dependencies]
+ink = { version = "=4.2.1", default-features = false }
+
+scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive"] }
+scale-info = { version = "2.6", default-features = false, features = ["derive"], optional = true }
+
+[dev-dependencies]
+drink = { path = "../../drink", features = ["session"] }
```
deuszx, line 16: Seems like we use this `session` feature every time - can we make it a default?

examples/mocking/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+    #[test]
+    fn call_mocked_message() -> Result<(), Box<dyn Error>> {
+        let mut session = Session::<MinimalRuntime>::new()?;
+
+        // Firstly, we create the mocked contract.
+        const RETURN_VALUE: (u8, u8) = (4, 1);
+        let mocked_contract =
+            ContractMock::new().with_message(CALLEE_SELECTOR, mock_message(|()| RETURN_VALUE));
+
+        // Secondly, we are deploy it, similarly to a standard deployment action.
```
deuszx:
```suggestion
        // Secondly, we deploy it, similarly to a standard deployment action.
```

examples/mocking/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+        let mock_address = session.mocking_api().deploy_mock(mocked_contract);
```
deuszx: Looking at the usage of the API here - I think `deploy` will be sufficient (i.e. no need for `_mock` suffix). We are in the `mocking_api` context.

drink/src/mock/mocking_api.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Deploy `mock` as a standard contract. Returns the address of the deployed contract.
+    fn deploy_mock(&mut self, mock: ContractMock) -> AccountIdFor<R>;
+
+    /// Mock part of an existing contract. In particular, allows to override real behavior of
+    /// deployed contract's messages.
+    fn mock_existing_contract(&mut self, mock: ContractMock, address: AccountIdFor<R>);
+}
+
+impl<R: Runtime> MockingApi<R> for Sandbox<R> {
+    fn deploy_mock(&mut self, mock: ContractMock) -> AccountIdFor<R> {
```
deuszx: As mentioned in the comment under the example - can we rename to `deploy`?

drink/src/lib.rs
```diff
@@ -77,7 +96,65 @@ impl<R: Runtime> Sandbox<R> {
...
+            // We intercept the call and return the result of the mock.
+            Some(mock) => {
+                let (selector, call_data) = input_data.split_at(4);
+                let selector: Selector = selector
+                    .try_into()
+                    .expect("Input data should contain at least selector bytes");
+
+                let result = mock
+                    .call(selector, call_data.to_vec())
+                    .expect("TODO: let the user define the fallback mechanism");
```
deuszx, line 150: At least to me the expect message is not actionable. `call` can, in particular, return `MessageNotFound` - which I imagine would be a common mistake. Can we match on that variant specifically and return something more understandable?

drink/src/lib.rs
```diff
@@ -77,7 +96,65 @@ impl<R: Runtime> Sandbox<R> {
...
+
+                let result: ExecResult = Ok(ExecReturnValue {
+                    flags: ReturnFlags::empty(),
```
deuszx: Should we set the flags to `Revert` if the call failed?

drink/src/mock.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Creates a new registry.
+    pub fn new() -> Self {
+        Self {
+            mocked_contracts: BTreeMap::new(),
+        }
+    }
+
+    /// Registers `mock` for `address`.
+    pub fn register(&mut self, address: AccountId, mock: ContractMock) {
+        self.mocked_contracts.insert(address, mock);
```
deuszx: [`BTreeMap::insert`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.insert) returns `Option<V>` - previous element if any. Can we return it here as well? Just in case it turns out in the future that someone needs it. I don't see any reason why we should not.

drink/src/lib.rs
```diff
@@ -35,6 +48,8 @@ pub type EventRecordOf<T> =
 /// A sandboxed runtime.
 pub struct Sandbox<R: Runtime> {
     externalities: TestExternalities,
+    mock_registry: Arc<Mutex<MockRegistry<AccountIdFor<R>>>>,
```
deuszx, line 52: Does it have to be `Arc<Mutex<_>>` if `MockRegistry` itself is basically that?

deuszx: Ah, I misread the `MockRegistry` with `MockExtension`.

.github/workflows/rust-checks.yml
pmikolajczyk41: lemme do it in the next PR (#63)

drink/src/lib.rs
pmikolajczyk41: nope:
```
the trait `Input` is not implemented for `[u8]`
```

that's awful, I agree :P

examples/mocking/Cargo.toml
pmikolajczyk41: new issue created - #64  (I want to make such change separately for transparency)

drink/src/mock.rs
pmikolajczyk41: :white_check_mark:

drink/src/lib.rs
pmikolajczyk41: you mean: instead of the `expect` above?

drink/src/lib.rs
pmikolajczyk41: but here we are returning directly to the pallet, so there's nothing smart we can do here now I'm afraid...

drink/src/mock/mocking_api.rs
pmikolajczyk41: :white_check_mark:

examples/mocking/lib.rs
pmikolajczyk41: :white_check_mark: (although it was more consistent with `deploy_contract` from `ContractsApi`

examples/mocking/lib.rs
pmikolajczyk41: :white_check_mark:

drink/src/mock/contract.rs
```diff
@@ -0,0 +1,64 @@
...
+    mock::{error::MockingError, MockedCallResult},
+};
+
+/// Alias for a 4-byte selector.
+pub type Selector = [u8; 4];
+/// An untyped message mock.
+///
+/// Notice that in the end, we cannot operate on specific argument/return types. Rust won't let us
+/// have a collection of differently typed closures. Fortunately, we can assume that all types are
+/// en/decodable, so we can use `Vec<u8>` as a common denominator.
```
deuszx, line 16: 🧠

drink/src/mock/mocking_api.rs
```diff
@@ -0,0 +1,61 @@
...
+        // ensure that the pallet will treat our mock just as a regular contract, until we actually
+        // call it.
+        let mock_bytes = wat::parse_str(DUMMY_CONTRACT).expect("Dummy contract should be valid");
+        let mock_address = self
+            .deploy_contract(
+                mock_bytes,
+                0,
+                vec![],
+                // We have to use a different account ID for each contract.
+                vec![self.mock_counter as u8],
```
deuszx, line 30: If you had used some random number generator, could we get rid of the `mock_counter`?

drink/src/runtime/pallet_contracts_debugging/runtime.rs
```diff
@@ -0,0 +1,79 @@
...
+    ) -> Vec<u8> {
+        // By default, do not intercept, continue with the standard procedure.
+        None::<()>.encode()
+    }
+}
+
+decl_extension! {
+    /// A wrapper type for the `InterceptingExtT` debug extension.
+    pub struct InterceptingExt(Box<dyn InterceptingExtT + Send>);
+}
```
deuszx, line 74: I wonder why those traits, extensions etc (code above) can't be part of the Substrate. The `CallInterceptor` is already there and abstractions above seem generally applicable.

drink/src/runtime/pallet_contracts_debugging/intercepting.rs
```diff
@@ -0,0 +1,25 @@
...
+        // supported).
+        let intercepting_result = contract_call_debugger::intercept_call(
+            contract_address.encode(),
+            matches!(*entry_point, ExportedFunction::Call),
+            input_data.to_vec(),
+        );
+
+        Decode::decode(&mut intercepting_result.as_slice()).expect("Decoding should succeed")
+    }
+}
```
deuszx, line 25: If the `ContractcallDebugger` was part of Substrate, alongside the `CallInterceptor`, there could be a generic implementation 
```rust
impl<T, R: Runtime> CallInterceptor<R> for T { ... }
```
since, unless I misunderstand sth (which is highly probable), there's nothing about `DrinkDebug` that makes it impossible to implement generically.

deuszx (approved):
Just come comments/questions.

deuszx (approved):
Just come comments/questions.

deuszx (approved):
Just come comments/questions.

drink/src/lib.rs
deuszx: No. When the `result` above is an error returned from the mock. I think that's what happens for RPC calls - i.e. dry-runs. I'm not sure if we should behave like that here as well but I'm just putting it up

drink/src/lib.rs
deuszx: So the `expect(...)` message will be "thrown" to the runtime, not the user?

drink/src/runtime/pallet_contracts_debugging/runtime.rs
pmikolajczyk41: they could be, but I'm worried that pushing these things to Parity will be very painful... I can ask them on Element about it

drink/src/mock/mocking_api.rs
pmikolajczyk41: yes we could losing determinism and repeatability in tests (or using some seed...); I don't have strong opinion on that

drink/src/lib.rs
pmikolajczyk41: but this will propagate to the user as a panic from the sandbox(externalities(runtime)))

pmikolajczyk41: and thus we can't survive the info about mock all the way... it would require some work on the pallet side (e.g. new error variant)

drink/src/runtime/pallet_contracts_debugging/intercepting.rs
pmikolajczyk41: yeah, that's right; but at least for a few weeks it has to reside here, until they agree to accept this code in pallet and they review my PR xD

deuszx: Sure, understand it. It was also a question to clear up _my_ understanding about the system and where each piece (could) lie.

drink/src/lib.rs
deuszx: OK, let's leave it as is then.

drink/src/lib.rs
pmikolajczyk41: done, great spot!

drink/src/mock/mocking_api.rs
pmikolajczyk41: discussed offline: tbd in next PR

PR #68, opened 2023-10-19
Enable loading contract bundles

piotrMocz:
As simple as the title suggests. Instead of loading the `.wasm` blob and the `.json` metadata separately, you can now load the whole `.contract` file using one of:
```rust
ContractFile::load("some/path/to/your.contract")
local_contract_file!()
```
And pass them to the methods that work with bundles, which would be one of:
* `deploy_contract`
* `deploy_contract_and`
* `load_contract`
* `load_contract_and`

So, all in all, what you do in your tests is:
```rust
use drink::{
    local_contract_file,
    session::{ContractFile, Session}};

let contract = ContractFile::load("path/to/your.contract");
Session::<MinimalRuntime>::new()?.deploy_contract_and(contract, "new", NO_ARGS, vec![], None) /* ... */

// or:

Session::<MinimalRuntime>::new()?.deploy_contract_and(local_contract_file!(), "new", NO_ARGS, vec![], None) /* ... */
```

### DIFF
--- a/drink/Cargo.toml
+++ b/drink/Cargo.toml
@@ -10,7 +10,7 @@ version.workspace = true
 description = "Minimal sufficient architecture that allows for a fully functional ink! contract development"
 
 [dependencies]
-contract-metadata = { workspace = true }
+contract-metadata = { workspace = true, optional = true}
 contract-transcode = { workspace = true, optional = true }
 frame-metadata = { workspace = true }
 frame-support = { workspace = true }
@@ -25,7 +25,7 @@ sp-externalities = { workspace = true }
 sp-io = { workspace = true }
 sp-runtime-interface = { workspace = true }
 
-serde_json = { workspace = true }
+serde_json = { workspace = true, optional = true }
 scale-info = { workspace = true }
 thiserror = { workspace = true }
 wat = { workspace = true }
@@ -36,5 +36,5 @@ default = [
     "std",
     "session"
 ]
-session = ["contract-transcode"]
+session = ["contract-metadata", "contract-transcode", "serde_json"]
 std = []
--- a/drink/Cargo.toml
+++ b/drink/Cargo.toml
@@ -10,6 +10,7 @@ version.workspace = true
 description = "Minimal sufficient architecture that allows for a fully functional ink! contract development"
 
 [dependencies]
+contract-metadata = { workspace = true }
 contract-transcode = { workspace = true, optional = true }
 frame-metadata = { workspace = true }
 frame-support = { workspace = true }
@@ -24,6 +25,7 @@ sp-externalities = { workspace = true }
 sp-io = { workspace = true }
 sp-runtime-interface = { workspace = true }
 
+serde_json = { workspace = true }
 scale-info = { workspace = true }
 thiserror = { workspace = true }
 wat = { workspace = true }
--- a/drink/src/session.rs
+++ b/drink/src/session.rs
@@ -17,7 +17,7 @@ use crate::{
 };
 
 mod bundle;
-pub use bundle::ContractBundle;
+pub use bundle::ContractFile;
 pub mod error;
 mod transcoding;
 
@@ -97,8 +97,8 @@ pub const NO_ARGS: &[String] = &[];
 /// You can also work with `.contract` bundles like so:
 /// ```rust, no_run
 /// # use drink::{
-/// #   local_bundle,
-/// #   session::{ContractBundle, Session},
+/// #   local_contract_file,
+/// #   session::{ContractFile, Session},
 /// #   session::NO_ARGS,
 /// #   runtime::MinimalRuntime
 /// # };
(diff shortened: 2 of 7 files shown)


drink/Cargo.toml
```diff
@@ -10,6 +10,7 @@ version.workspace = true
 description = "Minimal sufficient architecture that allows for a fully functional ink! contract development"
 
 [dependencies]
+contract-metadata = { workspace = true }
```
pmikolajczyk41: this also should be optional (and enabled with `session` feature`)

drink/Cargo.toml
```diff
@@ -24,6 +25,7 @@ sp-externalities = { workspace = true }
 sp-io = { workspace = true }
 sp-runtime-interface = { workspace = true }
 
+serde_json = { workspace = true }
```
pmikolajczyk41: this one as well

drink/src/session.rs
```diff
@@ -395,3 +470,71 @@ impl<R: Runtime> Session<R> {
         self.sandbox.override_debug_handle(d);
     }
 }
+
+/// A struct representing the result of parsing a `.contract` bundle file.
```
pmikolajczyk41: I'd put all the code below (bundle-related) into a separate submodule in a `session/` folder - this file is long enough already :)

drink/src/session.rs
```diff
@@ -395,3 +470,71 @@ impl<R: Runtime> Session<R> {
...
+        })?;
+
+        let ink_metadata = serde_json::from_value(serde_json::Value::Object(metadata.abi))
+            .map_err(|e| {
+                SessionError::BundleLoadFailed(format!(
+                    "Failed to parse metadata from the contract file:\n{e:?}"
+                ))
+            })?;
+
+        let bytes = metadata
```
pmikolajczyk41: awful nit, but I would replace `bytes` and `ink_metadata` definition order (so that things are epsilon more local) :P

drink/src/session.rs
```diff
@@ -395,3 +470,71 @@ impl<R: Runtime> Session<R> {
...
+
+        let transcoder = Rc::new(ContractMessageTranscoder::new(ink_metadata));
+
+        Ok(Self { bytes, transcoder })
+    }
+
+    /// Load the `.contract` bundle (`bundle_file`) located in the `project_dir`` working directory.
+    ///
+    /// This is meant to be used predominantly by the `local_bundle!` macro.
+    pub fn local(project_dir: &str, bundle_file: String) -> Self {
```
pmikolajczyk41: this one doesn't have to be public, does it? we don't advertise it as a recommended API for the end user, so maybe we can make it private

pmikolajczyk41: oh, it has to be public, because of macro, right?

pmikolajczyk41 (approved):
glamorous contribution, thank you very much!

drink/src/session.rs
```diff
@@ -395,3 +470,71 @@ impl<R: Runtime> Session<R> {
...
+/// of the current project.
+#[macro_export]
+macro_rules! local_bundle {
+    () => {
+        drink::session::ContractBundle::local(
+            env!("CARGO_MANIFEST_DIR"),
+            env!("CARGO_CRATE_NAME").to_owned() + ".contract",
+        )
+    };
+}
```
deuszx: Pardon my ignorance but why does it have to be a macro and can't be a simple function?

pmikolajczyk41: bc you want it to be resolved in the target crate, not in drink :P

deuszx: but if it's used as a function in the body of the contract's test then it will resolve there, right?

pmikolajczyk41: not sure when `env!` will be resolved... @piotrMocz had problems with it

deuszx: Aye! That might be it.

drink/src/session.rs
deuszx: Can we rename `bundle_file` to `contract_file`? I think to users coming to use it "bundle" will not mean anything.

We could also verify that `contract_file` has `*.contract` extension.

deuszx (approved):

drink/src/session.rs
piotrMocz: @deuszx the reason I went with `bundle` instead of `contract` (especially in the name of the struct but here as well) was to make it very clear we're not talking about some deployed contract instance but just the result of parsing the `.contract` file. But maybe the name `ContractFile` would be also better than `ContractBundle`?

piotrMocz: @pmikolajczyk41 yeah, that was my reasoning here :)

drink/src/session.rs
piotrMocz: Right, so to the best of my knowledge (albeit very empirical) you have three options for loading the env vars in library code:
* `env!` inside a regular function: during the compilation of the library crate, in which case `CARGO_CRATE_NAME` will be `drink` -- not helpful
* `std::env::var`: loading at runtime. `CARGO_CRATE_NAME` won't be defined because it's set only for the duration of the compilation
* `env!` inside a macro, at which point it will be substituted during the compilation of the project that's using said library crate -- bingo :)

If I'm missing something let me know!

drink/src/session.rs
deuszx: > @deuszx the reason I went with `bundle` instead of `contract` (especially in the name of the struct but here as well) was to make it very clear we're not talking about some deployed contract instance but just the result of parsing the `.contract` file. But maybe the name `ContractFile` would be also better than `ContractBundle`?

`ContractBundle` for the struct is OK, I was only worried about the arg's name.

PR #70, opened 2023-10-19
#[drink::test] macro

### LINKED ISSUE #62: Build contract automatically
with build script instead of requiring that `cargo contract build` is invoked before `cargo test`

pmikolajczyk41:
Closes #62

In this PR we introduce `#[drink::test]` macro. It builds contract(s) for a testcase, similarly to the `ink_e2e::test`. 

Follow-up: Provide bundles for all built contracts to a testcase.


drink/test-macro/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+mod ir;
+
+type SynResult<T> = Result<T, syn::Error>;
+
+/// Defines a drink!-based test. *WARNING: THE MACRO DOES NOTHING YET (MISSING CONTRACT BUILDING
+/// FUNCTIONALITY). TO BE ADDED WITHIN A FEW DAYS.*
+///
+/// # Requirements
+///
+/// - `drink` crate should be available in the target crate's dependencies (at the path `::drink`).
```
piotrMocz: I feel like this should come with an example ;)

piotrMocz (approved):
LGTM

drink/test-macro/src/lib.rs
pmikolajczyk41: no longer required :)

drink/test-macro/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+///
+/// # Requirements
+///
+/// - You mustn't import `drink::test` in the scope, where the macro is used. In other words, you
+/// should always use the macro only with a qualified path `#[drink::test]`.
+///
+/// # Impact
+///
+/// This macro will take care of building all needed contracts for the test. The building process
+/// will be executed during runtime. This means that the test will take longer to execute if the
```
deuszx:
```suggestion
/// will be executed during compile-time. This means that the test will take longer to execute if the
```
I hope?

drink/test-macro/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+/// contracts are not already built.
```
deuszx: ANd this sentence can/will be dropped.

pmikolajczyk41: yup, removed

drink/test-macro/src/lib.rs
pmikolajczyk41: right, fixed

drink/test-macro/src/ir.rs
```diff
@@ -0,0 +1,82 @@
+use darling::{ast::NestedMeta, FromMeta};
+use proc_macro2::TokenStream as TokenStream2;
+use syn::ItemFn;
+
+/// The macro arguments (available configuration).
```
deuszx: Arguments or attributes? :) Maybe I'm missing the nomenclature but we're passing the `attr` arg to `MacroArgs::from_list` - is it OK to use either `Args` or `Attr` everywhere?

pmikolajczyk41: kinda... I will unify it though

drink/test-macro/src/codegen.rs
```diff
@@ -0,0 +1,30 @@
...
+    // a bit.
+    let fn_signature = &item_fn.sig;
+    let fn_body = &item_fn.block;
+    let fn_vis = &item_fn.vis;
+    let fn_attrs = &item_fn.attrs;
+
+    // Prepare the code responsible for building the contracts.
+    let _manifests = ir.manifests();
+    let _debug_mode = ir.compile_in_debug_mode();
+    let build_contracts = quote! {};
```
deuszx: This will be where we build additional `manifest`-sourced deps? Since the "project" dependencies (defined in `Cargo.toml`) will be built earlier.

drink/test-macro/src/ir.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    #[test]
+    fn default_arguments() {
+        let attr = quote! {};
+        let ir = IR::try_from((attr, empty_function())).expect("failed to parse macro args");
+
+        assert_eq!(ir.manifests(), &Vec::<String>::new());
+        assert!(!ir.compile_in_debug_mode());
+    }
+}
```
deuszx: What we are missing (I think) is a test for function's attributes - i.e. ones that are NOT coming from `drink::test` macro.

drink/test-macro/src/contract_building.rs
```diff
@@ -0,0 +1,110 @@
...
+        network: Network::Online,
+        build_artifact: BuildArtifacts::All,
+        unstable_flags: UnstableFlags::default(),
+        optimization_passes: Some(OptimizationPasses::default()),
+        keep_debug_symbols: false,
+        lint: false,
+        output_type: OutputType::HumanReadable,
+        skip_wasm_validation: false,
+        target: Target::Wasm,
+    };
```
deuszx, line 102: I'm wondering - some of these params are definable in `Cargo.toml` - especially for the root package - and here we're "overwriting" them. That's probably fine since we are "optimising for drink!" but just something to be aware of.

deuszx (approved):

piotrMocz (approved):

drink/test-macro/src/contract_building.rs
pmikolajczyk41: https://github.com/Cardinal-Cryptography/drink/issues/72

PR #73, opened 2023-10-23
#[drink::contract_bundle_provider] macro

pmikolajczyk41:
Follow-up to #70 and #68.

We introduce a new macro `#[drink::contract_bundle_provider]` which provides a convenient way of obtaining contract bundles in drink testing.

This bundling provision wasn't done within `#[drink:test]` macro so as to avoid generating enums and implementations per-testcase. Now the user has to (although they can still get bytes/transcoders manually or use `ContractBundle` API) add one new type to the test module:
```rust
#[drink::contract_bundle_provider]
enum BundleProvider {}
```

which will be expanded to:

```rust
enum BundleProvider {
    Flipper,
    Counter,
}
impl BundleProvider {
    pub fn local() -> ::drink::DrinkResult<::drink::ContractBundle> {
        ::drink::ContractBundle::load(
            "/home/piomiko/Desktop/drink/examples/flipper/target/ink/flipper.contract",
        )
    }
    pub fn bundle(self) -> ::drink::DrinkResult<::drink::ContractBundle> {
        match self {
            BundleProvider::Flipper => ::drink::ContractBundle::load(
                "/home/piomiko/Desktop/drink/examples/flipper/target/ink/flipper.contract",
            ),
            BundleProvider::Counter => ::drink::ContractBundle::load(
                "/home/piomiko/Desktop/drink/examples/counter/target/ink/counter.contract",
            ),
        }
    }
}
```
(assuming that we are using it within `flipper` project and we have `counter` as a contract dependency in `Cargo.toml`)


examples/flipper/lib.rs
```diff
@@ -57,7 +59,7 @@ mod tests {
 
     #[drink::test]
     fn flipping() -> Result<(), Box<dyn Error>> {
-        let contract = local_contract_file!();
+        let contract = BundleProvider::Flipper.bundle()?;
```
deuszx, line 62: This will get red underlined squiggles?

pmikolajczyk41: it depends on the IDE - mine (Rust Rover) is able to resolve it and even give hints on variants / methods :sunglasses:

it's quite a lightweight macro and thus some decent Rust analyzers should get it

examples/mocking/lib.rs
```diff
@@ -37,27 +37,19 @@ mod proxy {
...
-            ContractMessageTranscoder::load(PathBuf::from("./target/ink/mocking.json"))
-                .expect("Failed to create transcoder"),
-        )
-    }
-
-    fn bytes() -> Vec<u8> {
-        fs::read("./target/ink/mocking.wasm").expect("Failed to find or read contract file")
-    }
+    #[drink::contract_bundle_provider]
+    enum BundleProvider {}
```
deuszx, line 52: Not a fan of the DX - it's too magical for me. Personally I prefer the "in your face" approaches.

pmikolajczyk41: you can't get 'in your face' approach baking such info in the type system... I can see 3 options here:
 - drink provides no support for bundles, i.e. you have to use `ContractBundle` API with string literals, and its your responsibility to care for paths
 - drink provides e.g. mapping from contract name / manifest file to a `ContractBundle` as additional testcase argument (so your tests actually get one argument from the sky), and you depend on string identifiers / paths, for which you are responsible
 - drink provides a typed source of bundles (this PR)

deuszx: The thing with this PR is that I type in code that means nothing if you look at it. What about, sth like this:
```rust
#[drink::test(
  flipper = "path/to/flipper",
  counter = "path/to/counter",
  ...
)]
fn test(bundles: HashMap<String, ContractBundle>) {
 // flipper and counter available here
}
```

Which is basically what you propose in 2)?

pmikolajczyk41: how does these paths relate to dependencies in `Cargo.toml`? it'd be not cool to specify them for every testcase (actually, macro already knows all the paths)

---

> means nothing if you look at it

means more if you know what it does - every information is implicit (like contract names)

deuszx: > how does these paths relate to dependencies in Cargo.toml? it'd be not cool to specify them for every testcase (actually, macro already knows all the paths)

Hmm, good point. It might overwrite deps in `Cargo.toml`. But also it allows for adding ones that are not defined there. 

> means more if you know what it does - every information is implicit (like contract names)

Exactly. And this is what I don't like about it. 

I'm not gonna block the PR over it ofc.

drink/test-macro/src/lib.rs
```diff
@@ -62,3 +63,78 @@ fn test_internal(_attr: TokenStream2, item: TokenStream2) -> SynResult<TokenStre
...
+/// # Impact
+///
+/// This macro is intended to be used as an attribute of some empty enum. It will build all
+/// contracts crates (with rules identical to those of `#[drink::test]`), and populate the decorated
+/// enum with variants, one per built contract.
+///
+/// If the current crate is a contract crate, the enum will receive a method `local()` that returns
+/// the contract bundle for the current crate.
+///
+/// Besides that, the enum will receive a method `bundle(self)` that returns the contract bundle
```
deuszx, line 87:
```suggestion
/// Otherwise, the enum will receive a method `bundle(self)` that returns the contract bundle
```
?

pmikolajczyk41: nope, it's the `local` function which is an addition - `bundle` must be present to expose all available bundles

drink/test-macro/src/bundle_providing.rs
```diff
@@ -0,0 +1,85 @@
...
+            );
+        }
+
+        Self {
+            root_contract_name,
+            bundles,
+        }
+    }
+
+    pub fn generate_bundle_providing(&self, enum_item: ItemEnum) -> TokenStream2 {
```
piotrMocz: This is extremely nitpicky but I'd prefer `s/providing/provision/g` :smile:

piotrMocz (approved):
I think this is great! I'd say it still may have the potential to make it even more user-friendly but we can definitely discuss this after merging of this one (which I think is already a great step)

drink/test-macro/src/bundle_providing.rs
pmikolajczyk41: sure, done!

PR #74, opened 2023-10-26
Improve README

pmikolajczyk41:


UladAleph (commented):
I think it is generally good. I like the links to resources for docs. As well as the general description and separation of cli readme.

Do you think adding a short description of lib's modules and core functionality could be good as well? Possibly separately, just like the cli doc as well

pmikolajczyk41:
imho adding additional description for lib's modules is useful only for the ones, who want to contribute to the code itself

regarding public API, for now it is documented (cargo linter guards that), so that we don't need to describe it outside code

UladAleph (approved):

PR #76, opened 2023-10-27
Fix CI. Make `call` and `call_with_address` return decoded results

pmikolajczyk41:
CI:
 - Makefile should fail if any example fails
 - do not treat `target` directory as an example

Session:
 - make `call` and `call_with_address` return decoded results (`Vec<u8>` is useless)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: run build lint test_examples clean help
 
 EXAMPLES = ./examples
-EXAMPLES_PATHS := $(shell find $(EXAMPLES) -mindepth 1 -maxdepth 1 -type d)
+EXAMPLES_PATHS := $(shell find $(EXAMPLES) -mindepth 1 -maxdepth 2 -name Cargo.toml -exec dirname "{}" \;)
 EXAMPLES_TARGET = ./examples/target
 
 run: ## Run the project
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -17,8 +17,9 @@ lint: ## Run the linter
 test_examples: ## Run tests for the examples
 	@mkdir -p $(EXAMPLES_TARGET)
 	@for dir in $(EXAMPLES_PATHS); do \
-		echo "Processing $$dir" ; \
-		cargo test --quiet --manifest-path $$dir/Cargo.toml --release --target-dir $(EXAMPLES_TARGET); \
+		if [ $$dir = $(EXAMPLES_TARGET) ]; then continue; fi; \
+		echo "Processing $$dir"; \
+		cargo test --quiet --manifest-path $$dir/Cargo.toml --release --target-dir $(EXAMPLES_TARGET) || exit 1; \
```
deuszx, line 21: 🙊

Makefile
```diff
@@ -17,8 +17,9 @@ lint: ## Run the linter
 test_examples: ## Run tests for the examples
 	@mkdir -p $(EXAMPLES_TARGET)
 	@for dir in $(EXAMPLES_PATHS); do \
-		echo "Processing $$dir" ; \
-		cargo test --quiet --manifest-path $$dir/Cargo.toml --release --target-dir $(EXAMPLES_TARGET); \
+		if [ $$dir = $(EXAMPLES_TARGET) ]; then continue; fi; \
```
deuszx: Can you check whether changing `EXAMPLES_PATHS` to sth like
```
EXAMPLES_PATHS := $(shell find $(EXAMPLES)-mindepth 1 -maxdepth 1 -name Cargo.toml | xargs dirname)
```
help here? i.e. allow for getting rid of this little hack

deuszx (approved):

Makefile
pmikolajczyk41: had to change it a bit, but in the end hack was removed :)

PR #81, opened 2023-11-06
Expose the `fn execute_with`

kvinwang:
Phala's ink contract needs some extra setup. Like below code:
```Rust
        #[drink::test]
        fn deploy_and_call_http_get() -> Result<(), Box<dyn std::error::Error>> {
            use drink_pink_runtime::PinkRuntime;
            let checker_bundle = BundleProvider::local()?;
            let mut session = Session::<PinkRuntime>::new()?;
            session.execute_with(|| {
                PinkRuntime::setup_cluster().expect("Failed to setup cluster");
            });
            let checker =
                session.deploy_bundle(checker_bundle, "default", NO_ARGS, vec![], None)?;
            let (status, _body): (u16, String) = session.call_with_address(
                checker.clone(),
                "http_get",
                &["\"https://httpbin.org/get\""],
                None,
            )??;
            assert_eq!(status, 200);
     }
```

So, is it OK to export the function `execute_with`?

### DIFF
--- a/drink/src/chain_api.rs
+++ b/drink/src/chain_api.rs
@@ -81,6 +81,11 @@ pub trait ChainApi<R: Runtime> {
 
     /// Reset the events of the current block.
     fn reset_current_block_events(&mut self);
+
+    /// Execute the given closure with the inner externallities.
+    ///
+    /// Returns the result of the given closure.
+    fn execute_with<T>(&mut self, execute: impl FnOnce() -> T) -> T;
 }
 
 impl<R: Runtime> ChainApi<R> for Sandbox<R> {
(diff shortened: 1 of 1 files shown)


pmikolajczyk41 (changes requested):
generally, sure, I don't see any problems with that :ok_hand: 

however, please forgive me nitpicking, but I'd rather move this to `ChainApi` so that you can access this functionality from a `Session` object indirectly through `chain_api()` method

if that poses any implementation troubles, just let me know and I will help you with it :)

kvinwang:
> however, please forgive me nitpicking, but I'd rather move this to `ChainApi` so that you can access this functionality from a `Session` object indirectly through `chain_api()` method

Sounds good. I've moved it to `ChainApi`.

pmikolajczyk41 (approved):
thanks!

PR #84, opened 2023-11-08
Refactor Sandbox

pgherveou:
This PR does the following refactoring, to prepare further update that will make it easier to integrate Drink with a Network of chains connected through XCM:

Specifically it does the following:
- Use generics instead of hard coded types, so we can adapt e2e tests to various runtimes
- Refactor Sandbox so that it can be used with runtime that do not implement contracts (such as a Relay chain Runtime for example) 

For example token APIs only require the Sandbox to be generic over something that implement pallet_contracts

```rust
impl<R: pallet_balances::Config> Sandbox<R> {
    /// Add tokens to an account.
    ///
    /// # Arguments
    ///
    /// * `address` - The address of the account to add tokens to.
    /// * `amount` - The number of tokens to add.
    pub fn add_tokens(
        &mut self,
        address: AccountIdFor<R>,
        amount: BalanceOf<R>,
    ) -> Result<BalanceOf<R>, DispatchError> {
        self.execute_with(|| pallet_balances::Pallet::<R>::mint_into(&address, amount))
    }

    /// Add tokens to an account.
    ///
    /// # Arguments
    ///
    /// * `address` - The address of the account to add tokens to.
    /// * `amount` - The number of tokens to add.
    pub fn balance(&mut self, address: &AccountIdFor<R>) -> BalanceOf<R> {
        self.execute_with(|| pallet_balances::Pallet::<R>::free_balance(address))
    }
}

```


drink/src/chain_api.rs
```diff
@@ -1,158 +1,117 @@
...
+    /// # Arguments
+    ///
+    /// * `address` - The address of the account to add tokens to.
+    /// * `amount` - The number of tokens to add.
+    pub fn add_tokens(
+        &mut self,
+        address: AccountIdFor<R>,
+        amount: BalanceOf<R>,
+    ) -> Result<BalanceOf<R>, DispatchError> {
+        self.execute_with(|| pallet_balances::Pallet::<R>::mint_into(&address, amount))
```
pgherveou: ideally we keep existing pallet api names, but I did not want to break current API, if that's ok to make breaking change here and other apis, I am happy to do so as part of this PR

drink/src/session.rs
```diff
@@ -116,21 +111,21 @@ pub const NO_ARGS: &[String] = &[];
 ///     .deploy_bundle_and(contract, "new", NO_ARGS, vec![], None); /* ... */
 ///  # Ok(()) }
 /// ```
-pub struct Session<R: Runtime> {
+pub struct Session<R: Runtime + pallet_contracts::Config> {
```
pmikolajczyk41: I think it might be a convenient thing to introduce a new trait like `RuntimeWithContracts`, which would just capture this generic bound (it's already used in a couple of places)

drink/src/session.rs
```diff
@@ -184,18 +179,8 @@ impl<R: Runtime> Session<R> {
...
-
-    /// Returns a reference for basic contracts API.
-    pub fn contracts_api(&mut self) -> &mut impl ContractApi<R> {
-        &mut self.sandbox
-    }
-
-    /// Returns a reference for mocking API.
-    pub fn mocking_api(&mut self) -> &mut impl MockingApi<R> {
+    /// The underlying `Sandbox` instance.
+    pub fn sandbox(&mut self) -> &mut Sandbox<R> {
```
pmikolajczyk41: what is your reasoning for this? I have no strong opinion on that, but my initial intention was to express precisely what functionalities you are going to use

drink/src/runtime/mod.rs
```diff
@@ -20,33 +21,20 @@ pub use pallet_balances;
...
-pub trait Runtime:
-    frame_system::Config
-    + pallet_timestamp::Config
-    + pallet_balances::Config<Balance = u128>
-    + pallet_contracts::Config<Currency = pallet_balances::Pallet<Self>>
-{
+pub trait Runtime: frame_system::Config {
     /// Initialize the storage at the genesis block.
-    fn initialize_storage(_storage: &mut Storage) -> Result<(), String> {
-        Ok(())
```
pmikolajczyk41, line 33: why the default implementation is bad in this trait? I can imagine runtimes with empty work needed for some phases

drink/src/lib.rs
```diff
@@ -50,11 +47,10 @@ pub type EventRecordOf<T> =
     EventRecord<<T as frame_system::Config>::RuntimeEvent, <T as frame_system::Config>::Hash>;
 
 /// A sandboxed runtime.
-pub struct Sandbox<R: Runtime> {
+pub struct Sandbox<R: frame_system::Config> {
```
pmikolajczyk41: if we are abandoning runtime bounds, we should just be generic over account id for `mock_registry`

drink/src/chain_api.rs
```diff
@@ -1,167 +1,117 @@
...
+
 /// The runtime call type for a particular runtime.
 pub type RuntimeCall<R> = <R as frame_system::Config>::RuntimeCall;
 
-/// Interface for basic chain operations.
-pub trait ChainApi<R: Runtime> {
-    /// Return the current height of the chain.
-    fn current_height(&mut self) -> BlockNumberFor<R>;
-
+impl<R: Runtime> Sandbox<R> {
```
pmikolajczyk41: this block could go to the module with `Sandbox` itself

drink/src/chain_api.rs
pmikolajczyk41: we can break it, no problem with that (we will just have to bump version appropriately)

drink/src/chain_api.rs
```diff
@@ -180,10 +130,12 @@ mod tests {
...
-        let call = RuntimeCall::<MinimalRuntime>::Balances(
-            pallet_balances::Call::<MinimalRuntime>::transfer { dest, value },
-        );
-        sandbox.runtime_call(call, Some(MinimalRuntime::default_actor()).into())
+        sandbox.execute_with(|| {
+            RuntimeCall::<MinimalRuntime>::Balances(
+                pallet_balances::Call::<MinimalRuntime>::transfer { dest, value },
+            )
+            .dispatch(Some(MinimalRuntime::default_actor()).into())
+        })
```
pmikolajczyk41: out of curiosity - why is this better?

pmikolajczyk41 (commented):
I love the direction of this changes, that would finally allow to work freely with any runtime, not only for contract purposes, but for general interaction as well

if I understand the intention correctly, the idea is that:
 - `Sandbox` essentially wraps externalities and has a generic argument for a runtime, which can be actually anyhow configured
 - depending on this generic argument, `Sandbox` enables pallet-specific operations, but not through traits (`like ChainApi` and `ContractApi`), but rather through generic bounds

is that so?

drink/src/session.rs
pmikolajczyk41: oh, now I see that you are removing api traits at all

drink/src/chain_api.rs
pmikolajczyk41: or actually, we could have a module for sandbox instead of this `chain_api` module

pgherveou:
> I love the direction of this changes, that would finally allow to work freely with any runtime, not only for contract purposes, but for general interaction as well
> 
> if I understand the intention correctly, the idea is that:
> 
>     * `Sandbox` essentially wraps externalities and has a generic argument for a runtime, which can be actually anyhow configured
> 
>     * depending on this generic argument, `Sandbox` enables pallet-specific operations, but not through traits (`like ChainApi` and `ContractApi`), but rather through generic bounds
> 
> 
> is that so?

Yes. That's pretty much what Sandbox was doing already (as far as I understand).
I just moved the constraints around to make it more flexible and work with any Runtime as long as they implement the right config. We could re-introduce the traits and implement them for runtime that implement the right configs but I don't think it's necessary, simple generics bounds implementation should be enough here.

drink/src/mock/mocking_api.rs
```diff
@@ -1,12 +1,11 @@
 use crate::{
-    contract_api::ContractApi,
     mock::ContractMock,
     runtime::{AccountIdFor, Runtime},
     Sandbox, DEFAULT_GAS_LIMIT,
 };
 
 /// Interface for basic mocking operations.
-pub trait MockingApi<R: Runtime> {
+pub trait MockingApi<R: pallet_contracts::Config> {
```
deuszx: What's the reason for removing `ChainApi` and `ContractApi` traits but leaving this one?

drink/src/chain_api.rs
pgherveou: ah right, will move that where sandbox is defined

drink/src/lib.rs
pgherveou: with still need to be generic over some kind of runtime for the various implementation that will use the generic bound.

Ideally we move the mock stuff somewhere else since they are specific to contracts runtime

drink/src/runtime/mod.rs
pgherveou: Ah I can put that back, I felt that this could be removed since for now all implementations need to implement these fn explicitly

drink/src/session.rs
pgherveou: commented below. We can add these traits back if we implement them with the right constraints, but I don't think this add much value.

drink/src/chain_api.rs
pgherveou: not sure why I did that, should be the same, I will rollback

drink/src/runtime/mod.rs
pmikolajczyk41: yeah, I think we can restore it

drink/src/chain_api.rs
pgherveou: bumped to 0.7

pgherveou:
Addressed your comments

pmikolajczyk41 (approved):
thank you very much for this contribution!

examples/mocking/lib.rs
```diff
@@ -61,7 +61,7 @@ mod tests {
             ContractMock::new().with_message(CALLEE_SELECTOR, mock_message(|()| RETURN_VALUE));
 
         // Secondly, we deploy it, similarly to a standard deployment action.
-        let mock_address = session.mocking_api().deploy(mocked_contract);
+        let mock_address = session.sandbox().deploy(mocked_contract);
```
deuszx: I think it's a regression that users now have to even know about the `Sandbox`. It was supposed to be a low-level object and developers use high-level `session`.

pgherveou: yeah right this is indeed a bit confusing without the "mocking_api" name
will update

deuszx: It's not only about the name - name is just a "gate" into the `Sandbox` world. Of course, we can't stop users from using it, if they want, but I think the API should lead developer through the path of least resistance and suggest usage of simple constructs.

I'd prefer to leave the previous approach, if possible.

pgherveou: I think the issue here is that the mock utilities should not be tied to the Sandbox struct.
I restored the API and moved these props to the session in the last commit.
lmk what you think

Makefile
```diff
@@ -18,7 +18,7 @@ test_examples: ## Run tests for the examples
 	@mkdir -p $(EXAMPLES_TARGET)
 	@for dir in $(EXAMPLES_PATHS); do \
 		echo "Processing $$dir"; \
-		cargo test --quiet --manifest-path $$dir/Cargo.toml --release --target-dir $(EXAMPLES_TARGET) || exit 1; \
+		cargo test --manifest-path $$dir/Cargo.toml --release --target-dir $(EXAMPLES_TARGET) || exit 1; \
```
pgherveou: oups just me testing will rollback the --quiet

examples/mocking/lib.rs
deuszx: Looks good, thanks :)

pgherveou:
@deuszx @pmikolajczyk41 not sure what kind of approval is required to trigger CI here, you mind re-approving so we can merge this asap?

PR #85, opened 2023-11-17
Runtime creator

pmikolajczyk41:
We introduce a new macro `create_minimal_runtime`, which generates a minimal runtime with custom name and custom chain extension parameter. It can be used outside drink as well, but you also have to add `frame-system` and `frame-support` as dependencies (due to how `construct_runtime` macro works...).

This is a prerequisite for supporting chain extensions in drink.

### DIFF
--- a/drink/src/runtime/minimal.rs
+++ b/drink/src/runtime/minimal.rs
@@ -1,5 +1,6 @@
 #![allow(missing_docs)] // `construct_macro` doesn't allow doc comments for the runtime type.
 
+#[macro_export]
 macro_rules! create_minimal_runtime {
     ($name:ident) => {
         create_minimal_runtime!($name, ());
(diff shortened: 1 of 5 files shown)


drink/src/runtime/minimal.rs
```diff
@@ -1,142 +1,163 @@
...
-use crate::{
-    runtime::{pallet_contracts_debugging::DrinkDebug, AccountIdFor},
-    Runtime,
-};
+#[macro_export]
+macro_rules! create_minimal_runtime {
+    ($name:ident) => {
+        create_minimal_runtime!($name, ());
+    };
+    ($name:ident, $chain_extension: ty) => {
```
deuszx, line 8: Just FYI https://github.com/paritytech/ink/pull/1958

kroist (approved):

PR #87, opened 2023-11-21
Runtime interaction example

pmikolajczyk41:
Another example project, showing how one can interact with the runtime in drink.

### DIFF
--- a/examples/runtime-interaction/README.md
+++ b/examples/runtime-interaction/README.md
@@ -20,3 +20,5 @@ You can always switch from `Session` to `Sandbox` with:
     ...
     let sandbox = session.sandbox(); // `sandbox` has type `&mut Sandbox<Runtime>`
 ```
+
+A rule of thumb is: if you are working with contracts, use `Session`, otherwise use `Sandbox`.
--- a/examples/runtime-interaction/README.md
+++ b/examples/runtime-interaction/README.md
@@ -2,7 +2,7 @@
 
 This example shows how we can easily send transactions to a blockchain, just like with `subxt` or a similar client.
 
-This way we can leverage `drink!` to a library for a full-scope runtime simulation engine.
+This way we can leverage `drink!` to a full-scope runtime simulation engine.
 
 ## Running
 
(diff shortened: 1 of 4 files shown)


examples/runtime-interaction/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+This way we can leverage `drink!` to a library for a full-scope runtime simulation engine.
+
+## Running
+
+```bash
+cargo test --release
+```
+
+## `drink::Sandbox` vs `drink::Session`
```
UladAleph, line 13: Could you please elaborate a bit more on when users might choose one over the other. What are the key differences?

examples/runtime-interaction/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+        let call_object = RuntimeCall::Contracts(pallet_contracts::Call::remove_code {
+            code_hash: upload_result.code_hash,
+        });
+
+        sandbox
+            .runtime_call(call_object, Some(MinimalRuntime::default_actor()))
+            .expect("Failed to remove a contract");
+    }
+
+    const CONTRACT: &str = r#"
```
UladAleph, line 69: A bit bizarre I think. It is not clear for me from the example where this is being taken from. Do you think adding docs to show a way to get this from a contract could be good for the example?

pmikolajczyk41: added some explanation

examples/runtime-interaction/README.md
pmikolajczyk41: added some explanation

UladAleph (approved):
Solid work :)

examples/runtime-interaction/README.md
```diff
@@ -0,0 +1,27 @@
+# Runtime interaction
+
+This example shows how we can easily send transactions to a blockchain, just like with `subxt` or a similar client.
+
+This way we can leverage `drink!` to a library for a full-scope runtime simulation engine.
```
piotrMocz: leverage drink as a library?
I'm not sure but the above doesn't sound quite right to me :D

piotrMocz (approved):
Awesome :)

examples/runtime-interaction/README.md
pmikolajczyk41: good point: changed to `leverage 'drink!' to a full-scope runtime simulation engine.`

PR #89, opened 2023-11-23
Testing chain extension example

pmikolajczyk41:
Example showing how to inject arbitrary chain extension to `MinimalRuntime` and interact with it. Contains also minor improvement to the macro.

### DIFF
--- a/examples/chain-extension/src/chain_extension_ink_side.rs
+++ b/examples/chain-extension/src/chain_extension_ink_side.rs
@@ -6,7 +6,7 @@ pub trait StakingExtension {
     type ErrorCode = StakingExtensionErrorCode;
 
     /// Returns the number of the validators.
-    #[ink(extension = 41)]
+    #[ink(extension = 41, handle_status = false)]
     fn get_num_of_validators() -> u32;
 }
 
--- a/examples/chain-extension/src/chain_extension_ink_side.rs
+++ b/examples/chain-extension/src/chain_extension_ink_side.rs
@@ -1,0 +1,39 @@
+use ink::env::{chain_extension::FromStatusCode, DefaultEnvironment, Environment};
+
+/// Simple chain extension that provides some staking information.
+#[ink::chain_extension]
+pub trait StakingExtension {
+    type ErrorCode = StakingExtensionErrorCode;
+
+    /// Returns the number of the validators.
+    #[ink(extension = 41)]
+    fn get_num_of_validators() -> u32;
+}
+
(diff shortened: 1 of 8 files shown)


examples/chain-extension/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    create_minimal_runtime!(
+        RuntimeWithCE,
+        crate::chain_extension_runtime_side::StakingExtension
+    );
+
+    /// Test that we can call chain extension from ink! contract and get a correct result.
+    #[drink::test]
+    fn we_can_test_chain_extension() -> Result<(), Box<dyn std::error::Error>> {
+        let result: u32 = Session::<RuntimeWithCE>::new()?
+            .deploy_bundle_and(BundleProvider::local()?, "new", NO_ARGS, vec![], None)?
```
piotrMocz, line 57: A random suggestion: seeing this I kind of wish we had constants like `NO_ENDOWMENT` just for the reader's benefit. Not important for this PR though, just thinking out loud :)

piotrMocz (approved):
Amazing!

Took me a while to figure out how the chain extension argument was handled by the macro (because it was not changed and therefore collapsed in the viewer 😄)

examples/chain-extension/src/lib.rs
pmikolajczyk41: good idea - let me introduce these helpers in the next PR

examples/chain-extension/src/chain_extension_ink_side.rs
```diff
@@ -0,0 +1,39 @@
+use ink::env::{chain_extension::FromStatusCode, DefaultEnvironment, Environment};
+
+/// Simple chain extension that provides some staking information.
+#[ink::chain_extension]
+pub trait StakingExtension {
+    type ErrorCode = StakingExtensionErrorCode;
+
+    /// Returns the number of the validators.
+    #[ink(extension = 41)]
+    fn get_num_of_validators() -> u32;
```
UladAleph, line 10: A fairly minor stuff, but a chain extension method should return Result<T, E> by default.
I think with current setup it will counterintuitively still return the  Result<> type.
I believe, to return a u32 you'd need to add #[ink(handle_status = false)] macro on top of the method

https://use.ink/macros-attributes/chain-extension/#details-handle_status

pmikolajczyk41: yes, it was returning `Result` (see outdated `unwrap` in the contract's message); but I made things explicit in the last commit; thanks for catching this!

PR #90, opened 2023-11-30
Helper constants

pmikolajczyk41:
As requested here: https://github.com/inkdevhub/drink/pull/89#discussion_r1403807424


piotrMocz (approved):
I think the usage now looks really good!

PR #91, opened 2023-11-30
Update cargo-contract crate versions

SkymanOne:
As part of [ink! 5.0.0-rc](https://github.com/paritytech/ink/pull/2011) and [cargo-contract 4.0.0-rc](https://github.com/paritytech/cargo-contract/pull/1415) we need to update `drink` dependency in order to publish the updated version of `ink_e2e` crate.


pmikolajczyk41:
you probably have to add something like `wasm-instrument = { version = "0.4.0", features = ["sign_ext"] }` to fix compilation

ascjones:
Here is a solution to this very annoying problem: https://github.com/paritytech/cargo-contract/pull/1416.

deuszx:
Don't forget to bump drink! version as well.

pmikolajczyk41 (approved):

Cargo.toml
```diff
@@ -18,15 +18,15 @@ homepage = "https://github.com/Cardinal-Cryptography/drink"
...
 [workspace.dependencies]
 anyhow = { version = "1.0.71" }
 cargo_metadata = { version = "0.18.1" }
 clap = { version = "4.3.4" }
-contract-build = { version = "3.0.1" }
-contract-metadata = { version = "3.2.0" }
-contract-transcode = { version = "3.2.0" }
+contract-build = { version = "4.0.0-rc.1" }
+contract-metadata = { version = "4.0.0-rc.1" }
+contract-transcode = { version = "4.0.0-rc.1" }
```
deuszx, line 29: I just noticed that - I don't think the main release of `drink!` to crates should be done against release candidates. There are other projects that depend on `drink!` from crates.io and they might get silently broken now. 


Judging from the diff though - there aren't that many changes. Does it mean `ink5` is backwards comptible with `ink4` ?

PR #92, opened 2023-12-01
Accessing emitted events

pmikolajczyk41:
Implementing https://github.com/orgs/inkdevhub/projects/2/views/1?pane=issue&itemId=34267897

In order to simplify a bit `Session` object, we introduce `Record` type that collects all the results, returns and events that have been produced by contract interactions during a session.

### DIFF
--- a/drink/src/session.rs
+++ b/drink/src/session.rs
@@ -264,6 +264,7 @@ impl<R: RuntimeWithContracts> Session<R> {
             .encode(constructor, args)
             .map_err(|err| SessionError::Encoding(err.to_string()))?;
 
+        self.record.start_recording_events(&mut self.sandbox);
         let result = self.sandbox.deploy_contract(
             contract_bytes,
             endowment.unwrap_or_default(),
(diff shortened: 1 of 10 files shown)


drink/src/session.rs
```diff
@@ -286,14 +286,15 @@ impl<R: RuntimeWithContracts> Session<R> {
...
+        self.record.stop_recording_events(&mut self.sandbox);
 
         let ret = match &result.result {
             Ok(exec_result) if exec_result.result.did_revert() => {
                 Err(SessionError::DeploymentReverted)
             }
             Ok(exec_result) => {
                 let address = exec_result.account_id.clone();
-                self.deploy_returns.push(address.clone());
+                self.record.push_deploy_return(address.clone());
```
deuszx, line 297: Isn't address a `Copy` type?

deuszx (approved):
Code around event/result decoding is so similar to what there is in the `ink-wrapper` <-> `drink!` integration that it makes me wonder whether we're not taking it a bit too far on either side (drink! or ink-wrapper) and start duplicating the work.

piotrMocz (approved):

drink/src/session.rs
pmikolajczyk41: nope, we are relying on the trait bounds from `frame_system` which do not impose `Copy`

PR #97, opened 2024-01-31
Bump polkadot-sdk deps

pgherveou:


drink/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 //! a running node.
 
 #![warn(missing_docs)]
+#![recursion_limit = "1024"]
```
pmikolajczyk41: why do we need to increase limit?

Cargo.toml
```diff
@@ -18,7 +18,7 @@ homepage = "https://github.com/Cardinal-Cryptography/drink"
 license = "Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/Cardinal-Cryptography/drink"
-version = "0.8.6"
+version = "0.8.7"
```
pmikolajczyk41: the only thing I'm concerned, is that whether this bump is backwards compatible... if there is any chance that this will break existing things, I would bump to `0.9.0`

cc: @deuszx

pmikolajczyk41 (approved):

drink/src/lib.rs
pgherveou: We don't I think I just copied that from polkadot-sdk as I was initially struggling with the macro to compile.

pgherveou:
```suggestion
```

Cargo.toml
pgherveou: rolling this back keeping that for next PR

pgherveou:
@pmikolajczyk41 rolled back the version for now. Can we merge this so I can rebase follow up pr 
#98

PR #98, opened 2024-02-01
Implement Runtime with blanket impl

pgherveou:
Implement the Runtime trait for all struct that implements the required trait rather than doing it inside the macro.
This let us use existing runtime in our tests (as long as they satisfy  the requirements)
See added test.

### DIFF
--- a/drink/src/lib.rs
+++ b/drink/src/lib.rs
@@ -2,7 +2,6 @@
 //! a running node.
 
 #![warn(missing_docs)]
-#![recursion_limit = "1024"]
 
 mod bundle;
 pub mod errors;
--- a/drink/src/lib.rs
+++ b/drink/src/lib.rs
@@ -2,6 +2,7 @@
 //! a running node.
 
 #![warn(missing_docs)]
+#![recursion_limit = "1024"]
 
 mod bundle;
 pub mod errors;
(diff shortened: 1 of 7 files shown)


drink/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 //! a running node.
 
 #![warn(missing_docs)]
+#![recursion_limit = "1024"]
```
pgherveou:
```suggestion
```

drink/src/runtime.rs
```diff
@@ -39,14 +39,6 @@ pub trait Runtime: frame_system::Config {
...
     /// Default actor for the runtime.
     fn default_actor() -> AccountIdFor<Self>;
-
-    /// Metadata of the runtime.
-    fn get_metadata() -> RuntimeMetadataPrefixed;
-
-    /// Convert an account to an call origin.
-    fn convert_account_to_origin(
-        account: AccountIdFor<Self>,
-    ) -> <<Self as frame_system::Config>::RuntimeCall as Dispatchable>::RuntimeOrigin;
```
pgherveou, line 49: Removed these 2 as , (as far as I can tell) there are no easy way to implement them directly with the where clause. But they can be easily called on the concrete type directly if needed later on

pgherveou: ah looks like this might be problematic as it's used in the drink integration

PR #99, opened 2024-02-05
Update api to enable custom runtime

pgherveou:
This PR update how the API is structured: 

the `Runtime` trait is gone. The way it was structured,made it impossible to bring in an existing Runtime and pass it to a drink or ink! test. This PR moves these APIs to a `SandboxConfig` trait, and make the `Sandbox` generic over that trait.

This let us write tests like this

```rust
impl_sandbox_config!(
    struct MyConfig {
        runtime: MyRuntime;
        default_balance: DEFAULT_BALANCE;
        default_actor: ALICE;
    }
);

#[drink::test(config = MyConfig)]
fn test_flipping_with_custom_runtime(mut session: Session) -> Result<(), Box<dyn std::error::Error>> {
    // snip
}
``` 

this  `create_minimal_runtime` macro is still there, it just now call the new `impl_sandbox_config` macro internally to implement the `SandboxConfig` trait. 

I have an upcoming PR to ink! that will leverage these new API to call e2e tests against a custom runtime.
Specifically it will be used with the `pallet-contracts-mock-network` crate to test XCM host functions

```rust
use pallet_contracts_mock_network::{parachain::Runtime as ParachainRuntime, ALICE};

impl_sandbox_config!(
    struct ParachainSandbox {
        runtime: ParachainRuntime;
        default_balance: 1_000_000_000_000_000;
        default_actor: ALICE;
    }
);

/// Positive case scenario:
#[ink_e2e::test(backend(runtime_only(runtime = ParachainSandbox)))]
async fn transfer_with_xcm_execute_works<Client: E2EBackend>(
    mut client: Client,
) -> E2EResult<()> {
    // snip
}

### CODE REVIEW

File: drink/src/runtime.rs [hunk 1]
```diff
@@ -2,53 +2,12 @@
...
 pub use minimal::MinimalRuntime;
+pub mod pallet_contracts_debugging;
 
 /// The type of an account identifier.
 pub type AccountIdFor<R> = <R as frame_system::Config>::AccountId;
 
 /// The type of a hash.
 pub type HashFor<R> = <R as frame_system::Config>::Hash;
-
-/// A runtime to use.
```
pgherveou, line 18: This PR remove the Runtime trait and move the functionality to the `Sandbox` and the `SandboxConfig` trait. This let us bring a Runtime built in another crate and use it as the runtime of our Sandbox.

drink-cli/src/app_state/print.rs
```diff
@@ -1,4 +1,4 @@
-use drink::contract_api::decode_debug_buffer;
+use drink::contracts_api::decode_debug_buffer;
```
deuszx, line 1: This is not backwards compatible fix so we should bump minor, not patch.

My rule of thumb is: you bump patch when downstream consumers don't need to do anything.

Cargo.toml
```diff
@@ -18,7 +18,7 @@ homepage = "https://github.com/Cardinal-Cryptography/drink"
 license = "Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/Cardinal-Cryptography/drink"
-version = "0.8.6"
+version = "0.8.7"
```
pmikolajczyk41: we should bump to `0.9.0` as some items are missing now (like the `Runtime` trait)

examples/flipper/Cargo.toml
```diff
@@ -15,6 +15,12 @@ scale-info = { version = "2.6", default-features = false, features = ["derive"],
 [dev-dependencies]
 drink = { path = "../../drink" }
 
+# testing custom runtime
+frame-support = "28.0.0"
+frame-system = "28.0.0"
+pallet-contracts-mock-network = "3.0.0"
```
pmikolajczyk41, line 21: actually, this flipper example didn't have any particular reason, but now, we can rename the whole example to e.g. `different-runtimes` and make it a showcase how to test a single contract against different environments

drink-cli/src/app_state/print.rs
pgherveou: Sounds right, will bump minor instead

drink/src/session/record.rs
```diff
@@ -21,20 +21,23 @@ type ContractExecResult<R> = pallet_contracts::ContractExecResult<BalanceOf<R>,
 /// By `result` we mean the full result (enriched with some context information) of the contract
 /// interaction, like `ContractExecResult`. By `return` we mean the return value of the contract
 /// execution, like a value returned from a message or the address of a newly instantiated contract.
-pub struct Record<R: RuntimeWithContracts> {
+pub struct Record<Config: SandboxConfig>
```
pmikolajczyk41: can we leave `Record` generic over `R: pallet_contracts::Config`? so that we don't introduce `Config: SandboxConfig` anywhere in this module

drink/src/session.rs
```diff
@@ -154,7 +161,7 @@ impl<R: RuntimeWithContracts> Session<R> {
         Ok(Self {
             sandbox,
             mocks,
-            actor: R::default_actor(),
+            actor: Config::default_actor(),
```
pmikolajczyk41, line 164: is this the only reason that `Session` is generic over `SandboxConfig` instead of `pallet_contracts::Config`?

pmikolajczyk41 (commented):
can you please elaborate on why the previous API was inconvenient? the difference is quite subtle

drink/src/session/record.rs
pgherveou: done I think I could not do it because of the recording APIs I moved that to the session

pgherveou:
> can you please elaborate on why the previous API was inconvenient? the difference is quite subtle

With the current API we expect that our (frame) runtime also implement `drink::Runtime`. We can do that because we create the runtime with the macro, so it's local to the crate where we use it.

If instead we try to bring in a (frame) Runtime that is declared in another crate. Then because of the orphan rule we won't be able to implement the foreign trait (drink::Runtime) on our (foreign frame) runtime.

To make this use case work,  this change [compose](https://github.com/inkdevhub/drink/blob/c8e6464a25bcbc94d83b3c126b53e67cb556d243/drink/src/sandbox/sandbox_config.rs#L11-L13) drink::Runtime inside a new `Config` trait, instead of using [trait inheritance](https://github.com/inkdevhub/drink/blob/42556188eade8eaa103c32eb23b56570f86d7621/drink/src/runtime.rs#L19)

drink/src/session.rs
pgherveou: mostly because the session now holds a sanbox that is generic over the SandboxConfig trait

examples/flipper/Cargo.toml
pgherveou: flipper is how we call "hello world" in ink! I This is just adding another test to the list, I think it's fine to keep it named like this. Happy to rename it though if you feel strongly about this.

pgherveou:
For reference this is how I am using this PR in ink!

https://github.com/paritytech/ink/blob/b0762ce5c066d66ffb9f8b27ff58ae9a2e9a1974/crates/e2e/src/drink_client.rs#L463-L479
```rust
/// Exposes preset sandbox configurations to be used in tests.
pub mod preset {
    use drink::impl_sandbox_config;
    use pallet_contracts_mock_network::{
        parachain::Runtime as ParachainRuntime,
        ALICE, INITIAL_BALANCE
    };

    drink::impl_sandbox_config!(
        #[doc = "A Sandbox configuration for the parachain runtime of pallet-contracts-mock-network."]
        pub struct ParachainConfig {
            runtime: ParachainRuntime;
            default_balance: INITIAL_BALANCE;
            default_actor: ALICE;
        }
    );
}
```

pmikolajczyk41 (approved):
'orphan rule' is the thing I was missing - thank you for pointing this, as well as for the contribution!

PR #105, opened 2024-02-22
Rename & update SandboxConfig trait

pgherveou:
Follow up from #99 

I realized that the current design was not working as I wanted.
Specifically, when working with parachain and `xcm_simulator::TestExt`, I need to use the existing externalities, and 
the custom [`execute_with`](https://paritytech.github.io/polkadot-sdk/master/pallet_contracts_mock_network/trait.TestExt.html#method.execute_with) to dispatch messages properly.

- The Sandbox is now created inside the macro (renamed `create_minimal_sandbox`) along with the runtime. The default setup now uses the `MinimalSandbox`  created through the macro, instead of  (the now deleted) `Sandbox` struct.

- The trait `SandboxConfig` is renamed `Sandbox` and exposes these two new APIs
```rust
    fn execute_with<T>(&mut self, execute: impl FnOnce() -> T) -> T;
    fn dry_run<T>(&mut self, action: impl FnOnce(&mut Self) -> T) -> T;
```

- `fn initialize_storage` is removed since this can be done when the Sandbox is created, and can be customized there instead of being part of the trait.
- The api in `sandbox/*_api.rs` are updated to be traits with blanket implementations for `T: Sandbox` instead of impl blocks. 

These updates let us create custom sandbox environment that have full control over the storage and execution.
An example of this, is the `MockNetworkSandbox` defined in [ink!](https://github.com/paritytech/ink/blob/ff584cb9fa406c04d30d39c494a3ac371bb28011/crates/e2e/src/drink_client.rs#L472-L536)
that uses under the hood the xcm simulator to test connected smart contract that need to interact with a Network of chains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.13.0]
+
+- Rework Sandbox API to better support custom Runtime
+
 ## [Unreleased]
 
 # [0.13.0]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ crossterm = { version = "0.26.0" }
 darling = { version = "0.20.3" }
 parity-scale-codec = { version = "3.6.9" }
 parity-scale-codec-derive = { version = "3.6.9" }
+paste = { version = "1.0.7" }
 proc-macro2 = { version = "1" }
 quote = { version = "1" }
 ratatui = { version = "0.21.0" }
--- a/drink-cli/src/app_state/mod.rs
+++ b/drink-cli/src/app_state/mod.rs
@@ -1,7 +1,7 @@
 use std::{env, path::PathBuf};
 
 pub use contracts::{Contract, ContractIndex, ContractRegistry};
-use drink::{runtime::MinimalRuntime, session::Session, SandboxConfig, Weight, DEFAULT_GAS_LIMIT};
+use drink::{runtime::MinimalSandbox, session::Session, Sandbox, Weight, DEFAULT_GAS_LIMIT};
 use sp_core::crypto::AccountId32;
 pub use user_input::UserInput;
 
@@ -23,7 +23,7 @@ impl Default for ChainInfo {
     fn default() -> Self {
         Self {
             block_height: 0,
-            actor: MinimalRuntime::default_actor(),
+            actor: MinimalSandbox::default_actor(),
             gas_limit: DEFAULT_GAS_LIMIT,
         }
     }
@@ -69,7 +69,7 @@ impl Default for UiState {
 }
 
 pub struct AppState {
-    pub session: Session<MinimalRuntime>,
+    pub session: Session<MinimalSandbox>,
     pub chain_info: ChainInfo,
     pub ui_state: UiState,
     pub contracts: ContractRegistry,
@@ -78,7 +78,7 @@ pub struct AppState {
 impl AppState {
     pub fn new(cwd_override: Option<PathBuf>) -> Self {
         AppState {
-            session: Session::new().expect("Failed to create drinking session"),
+            session: Session::default(),
             chain_info: Default::default(),
             ui_state: UiState::new(cwd_override),
             contracts: Default::default(),
--- a/drink/src/lib.rs
+++ b/drink/src/lib.rs
@@ -23,17 +23,15 @@ use pallet_contracts::{ContractExecResult, ContractInstantiateResult};
 #[cfg(feature = "session")]
 pub use session::mock::{mock_message, ContractMock, MessageMock, MockedCallResult, Selector};
 /// Export pallets that are used in the minimal runtime.
-pub use {frame_support, frame_system, pallet_balances, pallet_contracts, pallet_timestamp};
+pub use {
+    frame_support, frame_system, pallet_balances, pallet_contracts, pallet_timestamp, paste,
+    sp_externalities::Extension, sp_io::TestExternalities,
+};
 
-pub use crate::runtime::minimal::{self, MinimalRuntime};
-use crate::runtime::AccountIdFor;
+pub use crate::runtime::minimal::{self, MinimalSandbox};
 
 /// Alias for `frame-system`'s `RuntimeCall` type.
-pub type RuntimeCall<Runtime> = <Runtime as frame_system::Config>::RuntimeCall;
-
-/// Alias for `pallet-balances`'s Balance type.
-pub type BalanceOf<Runtime> =
-    <<Runtime as pallet_contracts::Config>::Currency as Inspect<AccountIdFor<Runtime>>>::Balance;
+pub type RuntimeCall<R> = <R as frame_system::Config>::RuntimeCall;
 
 /// Main result type for the drink crate.
 pub type DrinkResult<T> = std::result::Result<T, Error>;
@@ -43,6 +41,9 @@ pub type EventRecordOf<Runtime> = EventRecord<
     <Runtime as frame_system::Config>::RuntimeEvent,
     <Runtime as frame_system::Config>::Hash,
 >;
+
+type BalanceOf<R> =
+    <<R as pallet_contracts::Config>::Currency as Inspect<AccountIdFor<R>>>::Balance;
 
 /// Copied from pallet-contracts.
 pub type ContractInstantiateResultFor<Runtime> =
--- a/drink/src/runtime.rs
+++ b/drink/src/runtime.rs
@@ -1,10 +1,10 @@
-//! Module containing the [`Runtime`](Runtime) trait and its example implementations. You can use
-//! `drink` with any runtime that implements the `Runtime` trait.
+//! Module containing a [`MinimalSandbox`] that implements the [`crate::Sandbox`] trait.
+//! Also contains debugging utilities for the contracts pallet.
 
 pub mod minimal;
 pub mod pallet_contracts_debugging;
 pub use frame_metadata::RuntimeMetadataPrefixed;
-pub use minimal::MinimalRuntime;
+pub use minimal::MinimalSandbox;
 
 /// The type of an account identifier.
 pub type AccountIdFor<R> = <R as frame_system::Config>::AccountId;
--- a/drink/src/runtime/minimal.rs
+++ b/drink/src/runtime/minimal.rs
@@ -1,94 +1,88 @@
 #![allow(missing_docs)] // `construct_macro` doesn't allow doc comments for the runtime type.
 
-/// The macro will generate an implementation of `drink::SandboxConfig` for the given runtime type.
-#[macro_export]
-macro_rules! impl_sandbox_config {
-    (
-        $( #[ $attr:meta ] )*
-        $vis:vis struct $name:ident {
-            runtime: $runtime:tt;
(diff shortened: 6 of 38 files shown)


examples/flipper/lib.rs
```diff
@@ -71,34 +71,4 @@ mod tests {
 
         Ok(())
     }
-
```
pgherveou, line 74: This requires to implement the SandboxConfig trait manually, to keep things simple I am removing it here, the concrete example will live in ink!

pgherveou: For reference this will be pretty much used like this 
https://github.com/paritytech/ink/blob/3813349a9b9357dac68188518beaa21fe6664b8d/crates/e2e/src/drink_client.rs#L474-L538

pmikolajczyk41 (approved):
I'm not that big fan of the `thread_local` solution, but I don't want to block your work on this - let's merge it now and see if this causes any problems

also, two kind requests:

+ please bump version
+ in another PR, can you please add an example showcasing how you can use drink with custom runtime? even if it requires some manual work, I think it'd be good to have it covered here

drink/src/runtime/minimal.rs
```diff
@@ -1,65 +1,115 @@
...
-    (
-        $name:ident, $runtime:tt, $default_balance:expr, $default_actor:expr
-    ) => {
-        impl $crate::SandboxConfig for $name {
-            type Runtime = $runtime;
 
-            fn initialize_storage(storage: &mut $crate::frame_support::sp_runtime::Storage) -> Result<(), String> {
+    ($ext:ident, $name:ident, $runtime:tt, $default_balance:expr, $default_actor:expr) => {
+        thread_local! {
+            static $ext: ::std::cell::RefCell<$crate::TestExternalities> = ::std::cell::RefCell::new({
```
deuszx: IMO the addition of `thread_local!` and `RefCell` is a change for the worse. Previously the execution was predictable - the execution was linear and Rust borrow checker was making sure that we use the externalities properly. Now we've changed to internal mutability (note all the changes from `&mut` to `&`) and slapped on the `thread_local!` which slows everything down. 

Can you explain to me why MUST do it this way and cannot do it any other way?

Also, the changes here are backwards incompatible. You've changed the API of public interfaces and structs.

deuszx (dismissed):
Please convince me that we need the `thread_local!` and `RefCell`.

drink/src/runtime/minimal.rs
pmikolajczyk41: probably a middle solution is to keep in `Sandbox` a `dyn Externalities` object...

pgherveou: Yes I have to agree that I am not a big fan of the thead_local use myself neither. This was the fastest way for me to test that the solution worked. Let me see if I can find another workaround.

pgherveou:
@deuszx @pmikolajczyk41 updated the initial fix.

To avoid using the hacky `RefCell` &  thread_local solution, I updated (one more time 😓) how everything is setup.

- The Sandbox is now created inside the macro (renamed `create_minimal_sandbox`) along with the runtime. So we use `MinimalSandbox` instead of `Sandbox` for the default setup.
- The trait `SandboxConfig` is renamed `Sandbox` and exposes these two new APIs
```rust
    fn execute_with<T>(&mut self, execute: impl FnOnce() -> T) -> T;
    fn dry_run<T>(&mut self, action: impl FnOnce(&mut Self) -> T) -> T;
```
`initialize_storage` is removed since this can be done when the Sandbox is created, and can be customized there instead of being part of the trait.
- The api in `sandbox/*_api.rs` are updated to be traits with blanket implementations for `T: Sandbox` instead of impl blocks. 

This needs another pass to make sure some doc string are still up to date, but hopefully this address your concern.

deuszx:
Thanks @pgherveou , I'll review it this week. Also, before I forget, remember to bump minor version of the crate.

pgherveou:
> Thanks @pgherveou , I'll review it this week. Also, before I forget, remember to bump minor version of the crate.

Gentle ping for review 🙏

Cargo.toml
```diff
@@ -18,9 +18,10 @@ homepage = "https://github.com/Cardinal-Cryptography/drink"
 license = "Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/Cardinal-Cryptography/drink"
-version = "0.12.0"
+version = "0.13.0"
```
pmikolajczyk41: please add a note in CHANGELOG.md

Cargo.toml
```diff
@@ -18,9 +18,10 @@ homepage = "https://github.com/Cardinal-Cryptography/drink"
...
 
 [workspace.dependencies]
+paste = { version = "1.0.7" }
```
pmikolajczyk41: nit: please adhere to the lexicographic sort

pmikolajczyk41: the same applies to the other `Cargo.toml`s

drink/src/runtime.rs
```diff
@@ -1,10 +1,9 @@
-//! Module containing the [`Runtime`](Runtime) trait and its example implementations. You can use
-//! `drink` with any runtime that implements the `Runtime` trait.
+//! Module containing a [`MinimalSandbox`] that implements the [`crate::Sandbox`] trait.
```
pmikolajczyk41, line 1: as well as debugging utilities from `pallet_contracts_debugging` module

drink/src/session.rs
```diff
@@ -428,34 +427,34 @@ where
...
         args: &[S],
-        endowment: Option<BalanceOf<Config::Runtime>>,
-    ) -> Result<MessageResult<T>, SessionError> {
+        endowment: Option<BalanceOf<T::Runtime>>,
+    ) -> Result<MessageResult<V>, SessionError> {
         self.call_internal(Some(address), message, args, endowment)
     }
 
-    fn call_internal<S: AsRef<str> + Debug, T: Decode>(
+    fn call_internal<S: AsRef<str> + Debug, V: Decode>(
```
pmikolajczyk41, line 507: +1 for s/T/V

drink/src/sandbox/timestamp_api.rs
```diff
@@ -1,42 +1,52 @@
...
 /// Generic Time type.
 type MomentOf<R> = <R as pallet_timestamp::Config>::Moment;
 
-impl<Config: SandboxConfig> Sandbox<Config>
+/// Timestamp API for the sandbox.
+pub trait TimestampAPI<T: Sandbox>
 where
-    Config::Runtime: pallet_timestamp::Config,
+    T: Sandbox,
+    T::Runtime: pallet_timestamp::Config,
```
pmikolajczyk41: I'm wondering whether an associated type would be better here than a generic trait... My feeling is that you want to have a single specified `T: Sandbox` for the trait implementation, rather than be generic over any sandbox... No strong opinion here though

pmikolajczyk41 (approved):

drink/src/sandbox/timestamp_api.rs
pgherveou: Good point this make the trait agnostic of the Sandbox as well, switching to it

pgherveou:
@deuszx  you are still blocking this, would be nice if you could remove the "change requested" or delegate your review to someone else?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.13.0]
+
+- Rework Sandbox API to better support custom Runtime
```
deuszx, line 10: I know it's not your doing but `0.12.1` is now released so can we move the headers around so that only `0.13.0` is unreleased?

drink-cli/src/app_state/mod.rs
```diff
@@ -69,7 +69,7 @@ impl Default for UiState {
 }
 
 pub struct AppState {
-    pub session: Session<MinimalRuntime>,
+    pub session: Session<MinimalSandbox>,
```
deuszx, line 72: I think it's kinda weird that `Session` is now parameterised with the `Sandbox` - it encloses the sandbox but the parameter type should be something that was the common denominator for both - session and sandbox. 

Cannot point my finger to anything specific, so this won't be a blocker, but I just have a gut feeling that previous typing was more "appropriate".

drink/src/lib.rs
```diff
@@ -23,17 +23,18 @@ use pallet_contracts::{ContractExecResult, ContractInstantiateResult};
...
+pub use crate::runtime::minimal::{self, MinimalSandbox};
 
 /// Alias for `frame-system`'s `RuntimeCall` type.
-pub type RuntimeCall<Runtime> = <Runtime as frame_system::Config>::RuntimeCall;
+pub type RuntimeCall<R> = <R as frame_system::Config>::RuntimeCall;
 
 /// Alias for `pallet-balances`'s Balance type.
-pub type BalanceOf<Runtime> =
-    <<Runtime as pallet_contracts::Config>::Currency as Inspect<AccountIdFor<Runtime>>>::Balance;
+pub type BalanceOf<R> = <R as pallet_balances::Config>::Balance;
```
deuszx: What w twist (https://github.com/inkdevhub/drink/pull/106/files#r1509471655). @pmikolajczyk41 , are you OK with this change? Since you made the previous one and had some arguments it was "more correct".

drink/src/lib.rs
```diff
@@ -44,13 +45,19 @@ pub type EventRecordOf<Runtime> = EventRecord<
     <Runtime as frame_system::Config>::Hash,
 >;
 
+type ContractBalanceOf<R> =
+    <<R as pallet_contracts::Config>::Currency as Inspect<AccountIdFor<R>>>::Balance;
```
deuszx: Ah, I see we have different types for "Balances" and "ContractBalances". This, again _just_, feels wrong - but I understand this comes from Substrate deps. I think it's a place where a _serious_ mistakes can happen since some tools (like ContractsUI for example) have a special handling for `Balance` type.

drink/src/lib.rs
```diff
@@ -44,13 +45,19 @@ pub type EventRecordOf<Runtime> = EventRecord<
...
+pub type ContractInstantiateResultFor<Runtime> = ContractInstantiateResult<
+    AccountIdFor<Runtime>,
+    ContractBalanceOf<Runtime>,
+    EventRecordOf<Runtime>,
+>;
 
 /// Copied from pallet-contracts.
 pub type ContractExecResultFor<Runtime> =
-    ContractExecResult<BalanceOf<Runtime>, EventRecordOf<Runtime>>;
+    ContractExecResult<ContractBalanceOf<Runtime>, EventRecordOf<Runtime>>;
```
deuszx: What if a contract makes a native token transfer which type is `BalanceOf` (not `ContractBalanceOf`)?

deuszx (commented):
As requested @pgherveou , I'm unblocking the PR (apologies for the holding off, I forgot that request changes block the merge) but I can't approve it with clear conscience - I don't have enough time to review it thoroughly at the moment but I know it's important for you so I'm deferring to @pmikolajczyk41 here.

drink/src/lib.rs
pgherveou: ah this can go away, I probably missed that in the merge

drink/src/lib.rs
pgherveou: should be the same, will get rid of the type dup

drink-cli/src/app_state/mod.rs
pgherveou: > it encloses the sandbox but the parameter type should be something that was the common denominator for both

The sandbox itself enclose the runtime so the common denominator still stands

PR #106, opened 2024-02-22
Dry running contract interactions

pmikolajczyk41:
Enhancing `Session` with `dry_run_deployment` and `dry_run_call`, accompanied with an example showcasing dry-run capabilities for both contract interaction and, in general, arbitrary runtime transaction.

I chose just a single version (but general enough) for both deploying and calling contract to put behind dry run to avoid bloating up `Session`'s API.

This should be merged after #104 and then the CI should be green.


drink/src/lib.rs
```diff
@@ -17,36 +17,51 @@ use std::sync::{Arc, Mutex};
...
+    runtime::{pallet_contracts_debugging::InterceptingExtT, AccountIdFor},
 };
 
 /// Alias for `frame-system`'s `RuntimeCall` type.
-pub type RuntimeCall<R> = <R as frame_system::Config>::RuntimeCall;
+pub type RuntimeCall<Runtime> = <Runtime as frame_system::Config>::RuntimeCall;
 
 /// Alias for `pallet-balances`'s Balance type.
-pub type BalanceOf<R> = <R as pallet_balances::Config>::Balance;
+pub type BalanceOf<Runtime> =
```
piotrMocz, line 35: why not stick to the previous definition? this one seems quite convoluted

piotrMocz (approved):

drink/src/lib.rs
pmikolajczyk41: the difference is from which pallet we assume is present... and in most places where we need this type alias we are working with just contracts pallet :| but I know it's not perfect :disappointed:

deuszx: My undestanding was that `Balance` type is "defined" by the `pallet_balances` - not `pallet_contracts`. That's why on Aleph Zero `Balance` has 12 digits and on other chains it might have a different number. 

Is `pallet_contracts`' `Balance` type somehow "tied" to that from `pallet_balances`? Or, differently, can it be the case that `pallet_contracts`'s `Balance` will differ from the one in `pallet_balances`?

pmikolajczyk41: it depends on the runtime configuration (i.e. is tied there); in 99.9% this will be the same type, but using the balance from `pallet_contracts` is, say, more abstract

ipapandinas (approved):

PR #112, opened 2024-03-06
Drink 0.8.7

### LINKED ISSUE #107: Drink 0.8.4+ broken.
As stated in https://github.com/inkdevhub/drink/pull/91#discussion_r1414578186 , that PR broke `0.8.4+` versions due to the incompatible changes made in the `4.0.0-rc1` releases of `cargo-*` packages. 

This means that downstream clients (like ink-wrapper-types) is now broken. 

I'd suggest releasing `0.8.7` which depends on `carge-*` packages at version `3.2.0`.

image

pmikolajczyk41:
Fixes #107 

1. Revert #91 (go back to `ink@4.3.0` and `contract_*@3.2.0`)
2. Port #92 feature over these reversion
3. Bump rust to 1.74

If common-amm and ink-wrapper works well with `0.8.7`, then `0.8.6` and `0.8.5` will be yanked

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - release-*
   pull_request:
 
 concurrency:
--- a/.github/workflows/rust-checks.yml
+++ b/.github/workflows/rust-checks.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - release-*
   pull_request:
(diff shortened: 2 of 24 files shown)


deuszx (approved):
Thanks for doing this. 

As for "git hygienic" - can you try with `get revert` first next time? I understand that it might require more work "afterwards" (ie in the "fix the revert commit" commit) but it's cleaner what the flow/reasoning was. I think.

PR #119, opened 2024-03-18
Move sandbox to drink_sandbox

pgherveou:
See https://github.com/paritytech/ink/pull/2158


Cargo.toml
```diff
@@ -41,19 +41,12 @@ serde_json = { version = "1.0" }
 syn = { version = "2" }
 thiserror = { version = "1.0.40" }
 wat = { version = "1.0.71" }
+ink_sandbox = { path = "/Users/pg/github/ink/crates/e2e/sandbox" }
```
pgherveou: TODO: fix once published

drink/src/lib.rs
```diff
@@ -4,54 +4,33 @@
...
 
 #[cfg(feature = "macros")]
 pub use drink_test_macro::{contract_bundle_provider, test};
 pub use errors::Error;
-use frame_support::traits::fungible::Inspect;
-pub use frame_support::{
-    sp_runtime::{AccountId32, DispatchError},
-    weights::Weight,
+pub use frame_support;
+pub use ink_sandbox::{
```
pmikolajczyk41, line 15: just `ink_sandbox::*`?

pmikolajczyk41 (approved):

drink/src/lib.rs
pgherveou: yeah I just thought that `api as sandbox_api` would make it a bit clearer

PR #121, opened 2024-04-02
Return error when call was reverted

pmikolajczyk41:


deuszx (commented):
Can you include some example? What's the context for using it? Isn't it equivalent to calling a method that returns `Result<T, Err>` and asserting it's `Err`?

pmikolajczyk41:
@deuszx https://github.com/Cardinal-Cryptography/ink-exercises/blob/main/exercises/1-drink-test/tests/lib.rs#L61

it's not that straitghtforward - if your message returned `Err(E)`, then the `call` method returns `Err(<Ok(Err(E))>.encode())`. And it's better to handle this here, in drink, then every time on the client side

drink/src/session.rs
```diff
@@ -463,6 +463,22 @@ where
...
+    pub fn call_and_expect_error<S: AsRef<str> + Debug, E: Debug + Decode>(
+        &mut self,
+        message: &str,
+        args: &[S],
+        endowment: Option<BalanceOf<T::Runtime>>,
+    ) -> E {
+        self.call_internal::<_, Result<(), E>>(None, message, args, endowment)
+            .expect_err("Call should fail")
+            .decode_revert::<Result<(), E>>()
+            .expect("Call should be reverted")
```
deuszx: Here we're loosing information - `decode_revert` returns `None` if the underlying error is different than `CallReverted`. So, if it's anything else - we will fail on this expect and not provide any context for the developer to fix it. 

Maybe, instead of returning `Option` from `decode_revert`, we could return a `Result` and on the `Err` side include the original error?

pmikolajczyk41: done

deuszx (approved):

Issue #140, opened 2025-04-17
fix: Replace `chain-extension` test with similar precompile test


chungquantin:
~~Current issue is the `Memory` trait in `pallet-revive` is required for the implementation of the chain extension but the `Memory` trait is in [`wasm` module which is not exported as public](https://paritytech.github.io/polkadot-sdk/master/src/pallet_revive/lib.rs.html#32). In the chain extension implementation of Pop Network, we forked the `pallet-revive` and modified it. Probably deprecated in the future with the introduction of precompiles.~~ 

Chain extensions are deprecated and replaced with precompiles in `pallet-revive`

PR #143, opened 2025-04-17
feat: uplift dependencies and update to support `ink-v6.0.0-alpha`

chungquantin:
The PR uplift all dependencies to use latest version of ink! crate (v6.0.0) and migrate all code logic to support `pallet-revive`. Key critical changes made: 
- With the removal of `type Debug` in `ChainExtension`, the feature to mock contract calls fail completely and requires alternative approach using tracing API to support the feature. 
- Migrate `examples/quick-start-with-drink` to test with tracing API instead of decoding `debug_message` (which is removed in revive).
- Disable test `examples/chain-extension` (chain extensions are deprecated in `pallet-revive`) and adds todo for a similar precompile replacement.

## Examples Tracking
- [x] `examples/contract-events`
- [x] `examples/dry-running`
- [x] `examples/flipper`
- [x] `examples/runtime-interaction`
- [x] `examples/multiple-contracts`
- [x] `examples/quick-start-with-drink`

## TO BE FIXED 
Aside from all TODOs and deprecated calls of `MockExtension`, below issues need to be resolved: 
- [ ] `examples/cross-contract-call-tracing`: Consider implement the alternative tracing approach with the new [Debugging Strategies](https://github.com/use-ink/ink/blob/master/integration-tests/public/debugging-strategies/lib.rs)
- [ ] `examples/mocking`: Need to implement alternative approach without `type Debug` in `ChainExtension`
- [ ] Issue: https://github.com/use-ink/drink/issues/140 |`examples/chain-extension`: chain extensions are deprecated and replaced with precompiles in pallet-revive ~~Current issue is the `Memory` trait in `pallet-revive` is required for the implementation of the chain extension but the `Memory` trait is in [`wasm` module which is not exported as public](https://paritytech.github.io/polkadot-sdk/master/src/pallet_revive/lib.rs.html#32). In the chain extension implementation of Pop Network, we forked the `pallet-revive` and modified it. Probably deprecated in the future with the introduction of precompiles.~~


examples/chain-extension/src/chain_extension_runtime_side.rs
chungquantin, line 1: Code does not work as expected because `Memory` trait is in a private module `wasm`. Which will be replaced with precompile soon.

chungquantin:
## Note to the reviewers: 

- This PR is quite big due to the merge with https://github.com/chungquantin/drink/pull/2 (which was reviewed and approved by two reviewers). Recommend reviewing [following these changes](https://github.com/use-ink/drink/pull/143/files/da6eaa15e4e0e4eec200f514c424c9d59a18affb..9f00ed6830e2ab5e41b14e89dc64563797c495eb).

- This PR introduces breaking changes to some drink functionalities: contract call intercepting (https://github.com/use-ink/drink/issues/144) & tracing API (https://github.com/use-ink/drink/issues/145) but the basic uses of drink still work as expected, see above working example contracts. 

- I think it makes more sense to resolve other issues in separate PRs and marks not working features as deprecated until we solve them.

drink/src/session/mock.rs
```diff
@@ -7,18 +7,18 @@ use std::collections::BTreeMap;
...
 /// A registry of mocked contracts.
-pub(crate) struct MockRegistry<AccountId: Ord> {
-    mocked_contracts: BTreeMap<AccountId, ContractMock>,
+pub(crate) struct MockRegistry {
+    mocked_contracts: BTreeMap<H160, ContractMock>,
     nonce: u8,
 }
 
-impl<AccountId: Ord> MockRegistry<AccountId> {
+impl MockRegistry {
```
chungquantin, line 21: This is no longer used. Running code local warns the message that this needs to be updated: 

```
warning: use of deprecated struct `session::mock::extension::MockingExtension`: No longer used due to pallet-revive removed `type Debug` for intercepting contract calls
```

examples/cross-contract-call-tracing/lib.rs
```diff
@@ -59,115 +57,116 @@ mod contract {
...
-            ],
-            NO_ENDOWMENT,
-        )??;
-
-        assert_eq!(value, 22);
-
-        Ok(())
-    }
-}
+// TODO: Deprecated due to `pallet-revive`. There is no `debug_message`.
```
chungquantin, line 60: #145

examples/quick-start-with-drink/lib.rs
```diff
@@ -116,31 +107,33 @@ mod tests {
...
-        // In particular, we can get the decoded debug buffer from the contract. The buffer is
-        // just a vector of bytes, which we can decode using the `decode_debug_buffer` function.
-        let decoded_buffer = &session.record().last_deploy_result().debug_message;
-        let encoded_buffer = decode_debug_buffer(decoded_buffer);
-
-        assert_eq!(encoded_buffer, vec!["Initializing contract with: `true`"]);
-
-        Ok(())
-    }
+    // TODO: Deprecated due to `pallet-revive`. There is no `debug_message`.
```
chungquantin, line 124: #145

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.81"
-components = ["rustfmt", "clippy"]
+channel = "stable-aarch64-apple-darwin"
```
davidsemakula:
```suggestion
channel = "1.86"
```

Specifying the current channel means Drink! will only compile on Apple Silicon.

I'm actually not sure why Drink! is requiring a specific toolchain like this, do you? If not we can look into that later. 

For now, ink! v6 requires `rustc >= 1.86`, so `1.86` seems like a reasonable choice before we make a decision on the rest (i.e. the relevance of specifying a channel this way).

examples/multiple-contracts/Cargo.toml
```diff
@@ -1,14 +1,16 @@
 [package]
 name = "multiple-contracts"
-authors = ["Cardinal", "Aleph Zero Foundation"]
+authors = ["Use Ink <ink@use.ink>"]
```
davidsemakula:
```suggestion
authors = ["Cardinal", "Aleph Zero Foundation", "Use Ink <ink@use.ink>"]
```
Let's keep the credit for the original authors 🙂

Cargo.toml
```diff
@@ -1,32 +1,26 @@
...
+members = ["drink", "drink/test-macro", "drink-cli"]
 
-exclude = [
-    "examples/",
-]
+exclude = ["examples/"]
 
 [workspace.package]
-authors = ["Cardinal"]
+authors = ["Use Ink <ink@use.ink>"]
```
davidsemakula:
```suggestion
authors = ["Cardinal", "Use Ink <ink@use.ink>"]
```
Ditto about authors.

examples/chain-extension/Cargo.toml
```diff
@@ -1,33 +1,35 @@
 [package]
 name = "chain-extension"
-authors = ["Cardinal", "Aleph Zero Foundation"]
+authors = ["Use Ink <ink@use.ink>"]
```
davidsemakula:
```suggestion
authors = ["Cardinal", "Aleph Zero Foundation", "Use Ink <ink@use.ink>"]
```
Ditto about authors.

examples/dry-running/Cargo.toml
```diff
@@ -1,13 +1,13 @@
 [package]
 name = "dry-running"
-authors = ["Cardinal", "Aleph Zero Foundation"]
+authors = ["Use Ink <ink@use.ink>"]
```
davidsemakula:
```suggestion
authors = ["Cardinal", "Aleph Zero Foundation", "Use Ink <ink@use.ink>"]
```
Ditto about authors.

examples/flipper/Cargo.toml
```diff
@@ -1,16 +1,13 @@
 [package]
 name = "flipper"
-authors = ["Cardinal", "Aleph Zero Foundation"]
+authors = ["Use Ink <ink@use.ink>"]
```
davidsemakula:
```suggestion
authors = ["Cardinal", "Aleph Zero Foundation", "Use Ink <ink@use.ink>"]
```
Ditto about authors.

examples/quick-start-with-drink/lib.rs
```diff
@@ -20,20 +15,16 @@ mod flipper {
     impl Flipper {
         #[ink(constructor)]
         pub fn new(init: bool) -> Self {
-            debug_println!("Initializing contract with: `{init}`");
```
davidsemakula, line 23: Let's just comment out the `debug_println!` lines and leave TODO comments for bringing them back when a new API is ready.

drink-cli/src/executor/mod.rs
```diff
@@ -52,13 +59,14 @@ pub fn execute(app_state: &mut AppState) -> Result<()> {
...
         }
-
         CliCommand::Build => contract::build(app_state),
         CliCommand::Deploy {
             constructor,
             args,
             salt,
-        } => contract::deploy(app_state, constructor, args, salt),
+        } => {
+            contract::deploy(app_state, constructor, args, Some(vec_u8_to_array_32(salt)));
```
davidsemakula, line 68: It's better to change the `CliCommand::Deploy` parser/definition to take the salt as `Option<[u8; 32]>` than doing this conversion here.

drink/src/session/mock/extension.rs
```diff
@@ -1,24 +1,26 @@
...
-    pallet_contracts_debugging::InterceptingExtT,
+    pallet_revive::chain_extension::ReturnFlags,
     session::mock::{MockRegistry, Selector},
 };
 
+#[allow(dead_code)]
+#[deprecated = "No longer used due to pallet-revive removed `type Debug` for intercepting contract calls"]
 /// Runtime extension enabling contract call interception.
-pub(crate) struct MockingExtension<AccountId: Ord> {
+pub(crate) struct MockingExtension {
```
davidsemakula: AFAICT, this is not part of the public interface, so a deprecation warnings seems unnecessary.

Probably better to just comment this out and add a TODO to replace it with an equivalent API when we figure that out.

.github/workflows/rust-checks.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install Rust toolchain
-        uses: Cardinal-Cryptography/github-actions/install-rust-toolchain@v1
+        uses: use-ink/github-actions/install-rust-toolchain@v1
         with:
           targets: wasm32-unknown-unknown
```
davidsemakula:
```suggestion
```

Do we still need this target?

davidsemakula (changes requested):

davidsemakula:
https://github.com/use-ink/drink/blob/94835ac3124cc02d556be1d9ff9fdf419fc24465/drink-cli/src/executor/contract.rs#L121-L139

@chungquantin Looks like this (and may be its callers) need to be update to work with `.polkavm` binaries.

rust-toolchain.toml
chungquantin: This is a mistake when I tested it on local, thanks for pointing it out. Resolved in latest commit.

drink-cli/src/executor/mod.rs
chungquantin: The problem is that clap only accepts `Vec`, it is not possible to use slice type for `CliCommand::Deploy`

chungquantin:
> fix(drink-cli): support polkavm and fix address

Fixed in https://github.com/use-ink/drink/pull/143/commits/4b54e6f92f9f95c93168de65fddcc2f605170d35

.github/workflows/rust-checks.yml
chungquantin: No we don't, I have resolved it https://github.com/use-ink/drink/pull/143/commits/06e6f99366ea54ceb0252b67be3db19bb1e442eb

drink/src/session/mock/extension.rs
chungquantin: Reverted changes for this in https://github.com/use-ink/drink/pull/143/commits/410a6a1c451dd8fef3d073ad8141f42da3a56810

davidsemakula (approved):
Thanks @chungquantin !

